PR #3, opened 2022-10-18
Quantum circuit converters

### LINKED ISSUE #2: Adding circuit plotting

madagra:
Convert a quantum circuit built with PyQ into Qiskit. This is useful for visualization for example.

Closes #2


Brogis1:
Nice 👍 Thanks a lot @madagra !

pyqtorch/converters/store_ops.py
```diff
@@ -0,0 +1,58 @@
...
+    def clear(self, *args, **kwargs):
+        """Clear the current circuit visualization"""
+        self.operations = []
+        self.nqubits = 0
+
+
+ops_cache = OpsCache()
+
+
+def storable(func):
```
awennersteen, line 35: Running this on every evaluation must be unnecessarily expensive

awennersteen (changes requested):
Further review done offline

pyqtorch/converters/store_ops.py
madagra: @awennersteen I changed the logic and avoid adding the decorator. Now the operation storage is performed only when requested, i.e. in the converter routine. Let me know if it is now acceptable for merge.

madagra:
@awennersteen It should be better now. The actual storage of the operations in the forward pass is only performed when requested in the converter function, otherwise, nothing is done. This also implies that one needs to be a bit careful when adding new operations; now, the routine for storing operations must be explicitly called when needed in the operation implementation.

awennersteen:
LGTM

awennersteen (approved):

PR #4, opened 2022-10-21
Aw/add typing

### LINKED ISSUE #1: Enforce consistent typing with Mypy
Type annotation should be fully implemented across the whole code. Make sure `mypy` run actually works without errors.

awennersteen:
Closes #1


pyqtorch/core/measurement.py
```diff
@@ -12,28 +12,37 @@
...
+from typing import Any
+
 import torch
+
 from pyqtorch.core.operation import X, Y, Z
 
-qubit_operators = {'X': X, 'Y': Y, 'Z': Z}
+"""
+TODO: we should create the operator tensor only once, on startup in the network
+"""
```
madagra, line 23: @awennersteen What do you mean exactly here? Do you want to initialize the predefined qubit operators at startup?

pyqtorch/core/utils.py
```diff
@@ -12,15 +12,20 @@
...
-import numpy as np
 from string import ascii_letters as ABC
+from typing import Any, Sized
 
-ABC_ARRAY = np.array(list(ABC))
+import numpy as np
+import torch
+from numpy.typing import NDArray
 
+ABC_ARRAY: NDArray = np.array(list(ABC))
```
madagra, line 22: @awennersteen I do not really get why we need such an array. Why can't we just use a standard Numpy array in its place?

pyqtorch/core/utils.py
```diff
@@ -101,34 +110,34 @@ def _apply_batch_gate(state, mat, qubits, N_qubits, batch_size):
...
-    new_state_indices[qubits] = mat_indices[0:len(qubits)]
+    new_state_indices[qubits] = mat_indices[0 : len(qubits)]
 
     state_indices = "".join(list(state_indices))
     new_state_indices = "".join(list(new_state_indices))
     mat_indices = "".join(list(mat_indices))
 
     einsum_indices = f"{mat_indices},{state_indices}->{new_state_indices}"
-    #print(einsum_indices)
+    # print(einsum_indices)
```
madagra, line 140: This can be removed.

pyqtorch/matrices_sparse.py
```diff
@@ -106,40 +119,47 @@ def generate_ising_from_graph(graph,
...
                 raise ValueError("'type_ising' must be in ['Z', 'N']")
 
     return H
 
 
-def general_hamiltonian(graph=None, alpha=None, beta=None, gamma=None, device='cpu'):
-    #alpha, beta, gamma: matrices of parameters (e.g. alpha_ij)
-    #connectivity_graph: which qubits are connected
-    #no 1 body terms
+@typing.no_type_check
```
madagra, line 132: @awennersteen Is this disabling type check for the whole function body? A cleaner solution could be to use the `# type: ignore` notation on the offending lines if they are not too many.

madagra (approved):
I have some very minor comments but overall looks good to me.

pyqtorch/core/utils.py
awennersteen: My only work here was only adding the typing, this was Slimane. 
I'm not sure what's the need but we can probably find out from Slimane.

pyqtorch/core/measurement.py
awennersteen: yeah,
but I later realized that it is probably best kept out of the PyQ library

madagra: @awennersteen Agreed.

pyqtorch/core/utils.py
madagra: @awennersteen Got it. Let's keep it as it is for now then.

PR #5, opened 2022-10-21
Aw/improve batching

awennersteen:
It seems more user friendly to separate the batched operations from the non batched ones.
The current implementation of the ansatz broadcasts so it should be fine? But would love to hear about more ideas for refactoring because I originally had more in mind.

I've also added some basic tests, which just confirms that the output of the batched network is the same as for the non-batched ones.
They should probably be refactored.

includes #4


tests/test_operations.py
```diff
@@ -0,0 +1,119 @@
...
+        self.network = ModuleList(network)
+        self.state = QuantumCircuit(n_qubits)
+        self.operator = total_magnetization
+
+    def forward(self, nx: torch.Tensor) -> torch.Tensor:
+        batch_size = len(nx)
+        state = self.state.init_state(batch_size=batch_size, device=nx.device)
+        for layer in self.network:
+            state = layer(state, nx)
+        return self.operator(state, self.n_qubits, batch_size)
```
madagra, line 55: @awennersteen Can you add these to the `conftest.py`? It will be quite useful to have them as fixtures for reuse in future tests.

tests/test_operations.py
```diff
@@ -0,0 +1,119 @@
...
+            state = batchedRY(x[:, 0], state, [i], self.n_qubits)
+        return state
+
+
+class TestNetwork(Module):
+    def __init__(self, network, n_qubits=4) -> None:
+        super().__init__()
+        self.n_qubits = n_qubits
+        self.network = ModuleList(network)
+        self.state = QuantumCircuit(n_qubits)
```
madagra, line 47: As a minor suggestion: Why not just initiate the state once in the constructor and then reuse it in the forward pass? Something like:
```
def __init__(...):
    ...
    self.init_state = QuantumCircuit.init_state(batch_size, device)
    ...
```

tests/test_operations.py
```diff
@@ -0,0 +1,119 @@
...
+        return self.operator(state, self.n_qubits, batch_size)
+
+
+# TODO: these are all the same test, would be better to parameterize a test
+def test_batched_network():
+    ansatz = AlternateLayerAnsatz(n_qubits=4, n_layers=4)
+    network = TestNetwork([TestFM(), ansatz])
+    batched_network = TestNetwork([TestBatchedFM(), ansatz])
+    # to ensure the parameters are the same
+    batch_size = 2
```
madagra, line 64: @awennersteen I do not see how this actually ensures that the parameters are the same. The quantum circuits are still initialized with different parameters as far as I understand.

madagra (commented):
Few small comments. It looks good to me. I think we can address some larger refactoring in a follow-up PR and align with @Slimane33 on that.

pyqtorch/core/__init__.py
```diff
@@ -10,4 +10,4 @@
 # distributed under the License is distributed on an "AS IS" BASIS,
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
-# limitations under the License.
\ No newline at end of file
+# limitations under the License.
```
madagra, line 13: @awennersteen It might be a good idea to import all the available operations (batched and not) to this `__init__.py` file to simplify importing.

tests/test_operations.py
awennersteen: I only create the `ansatz` onece, i.e. it has one set of params.
We don't update these in this example. This is what ensures that they are the same. I will add that

The feature maps don't have any nn.Parameters, only `x` dependencies. Hence no parameters are initialized.

tests/test_operations.py
awennersteen: Yeah

tests/test_operations.py
awennersteen: but then we'd have to put the batch size in the network definition. 
I don't see the need for that - especially in the tests.

tests/test_operations.py
madagra: @awennersteen I see, thanks.

tests/test_operations.py
awennersteen: FYI making them fixtures proper would have involved some refactoring that I will do in a later PR

awennersteen:
I'm merging to move on, and I've opened a few issues to split up the todo work.

PR #10, opened 2022-11-24
Improve package documentation

### LINKED ISSUE #9: Add documentation
Some documentation should be added for new users, and perhaps restructure the current example.

Some further in-code technical documentation should also be added. For example for batching methods, and make it explicit that you can operate with non-batched gates through broadcasting on a batched state.

aagentile:
Attempting to make the usage of functions in the repo more clear. Added one more example. 
I think the target should start with primitives in `core`, then we can keep going also for functions outside the core in a similar fashion. So far I started with the 
- [x] `operation`

but maybe it is a good idea to include before merging also 
- [ ] `circuit`
- [ ] `measurement`
- [ ] `utils`
- [x] `batched operations`

Closes #9.


madagra:
@aagentile Added documentation for all operations. I think we can merge this one and open a new PR for adding additional docstrings and ReadTheDocs deployment. Are you OK with this?

PR #21, opened 2023-02-22
Adding Hamiltonian evolution with diagonalization

### LINKED ISSUE #20: Hamiltonian_evolution with matrix diagonalization
Currently, to implement $`e^{-iHt}`$, the `hamiltonian_evolution` uses a Runge-Kutta method to integrate the Schrodinger equation from 0 to t. This requires several matrix-vector products for each step between 0 and t, and is very unstable, leading to completely wrong results for high enough t.

Instead, the time-evolution operator can be implemented exactly by diagonalization. For an Hermitian matrix H:
```math
H = VDV^\dagger
```
Where $`V`$ is the matrix of eigenvectors and $`D`$ is the diagonal matrix of eigenvalues. Then, we have that
```math
e^{-iHt} = Ve^{-iDt}V^\dagger
```
Since $`D`$ is diagonal the exponentiation is just the exponential of the entries, and can be efficiently recomputed for different values of $`t`$. Note that if the Hamiltonian is already diagonal from the start, then it's even better.

This is still a "brute force" approach, but will work for a general hamiltonian up to maybe 12-13 qubits on a laptop (8000 X 8000 matrix) and more on the cluster. Larger hamiltonians should not be fully diagonalized directly, and specific symmetries should be exploited.

jpmoutinho:
Adds alternative function to perform the hamiltonian evolution through exact matrix diagonalization. Includes also the batched version. Includes unit tests to guarantee both options give the same results for small times (for larger times, the runge-kutta diverges).

Closes #20


madagra (approved):
This looks good to me.

PR #24, opened 2023-03-14
Fix qucint pipeline

jpmoutinho:
Add coalesce() to diagonalize checks


jpmoutinho:
Hey @madagra is it possible to update the v0.1.4dev tag to go up to this PR? It fixes a few things with the previous implementation. I've looked it up but not sure what is the correct way to do it.

PR #25, opened 2023-03-22
pypi installation

arthurfaria:
Hey @awennersteen,

here it is :)

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,42 @@
+name: Upload Release Package to PyPI
+
+on: push
+
+jobs:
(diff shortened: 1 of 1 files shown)


awennersteen:
@Slimane33 
I think you said somewhere you had added the PYPI creds, I guess you only did this for the real pypi
Could you register on test pypi and add those creds too?

If the current once are called `PYPI_<something>`, you can call those `TEST_PYPI_<something>`

Slimane33:
@awennersteen it is done, the test token is named TEST_PYPI_API_TOKEN

nmheim:
closing this

PR #28, opened 2023-04-03
Cphase bug fix, new tests and code refactoring

dominikandreasseitz:
Fixing issue: https://github.com/pasqal-io/PyQ/issues/27
1.  Added fix for batchCphase gate 
2. Added test_batched_operations.py in tests/
3. Changed default hatch envs to include precommit hooks, changed test pipeline with hatch
4. Added precommit yml file
5. Code formatting


pyproject.toml
```diff
@@ -30,19 +30,26 @@ dependencies = [
...
+  "pytest-mypy",
+  "pytest-xdist",
+  "nbconvert",
+  "ipykernel",
+  "pre-commit",
+  "flake8",
+  "black",
+  "isort",
+  "mypy"
 ]
```
awennersteen, line 48: I think the old behavior is better . Users shouldn't have to install pytest, pre-commit etc

dominikandreasseitz: ok i see! i am not done with the PR yet. i just wanted to improve the overall code quality by adding these tools so a dev can easily use them

madagra: Yes, but what @awennersteen was pointing out is that it is better to keep them in a separate `tests` environment rather than adding them to the default. Then you can just do something like:
```
python -m hatch --env tests run pytest
```

pyproject.toml
```diff
@@ -54,3 +61,61 @@ ignore_errors = false
...
+    .git,
+    .venv,
+"""
+extend-ignore = """
+    E266,
+    E741,
+    E742,
+"""
+per-file-ignores = """i
+    qucint.blocks/manipulate.py: F401
```
nmheim, line 85: remnant of copying from `qucint` :)

nmheim (approved):

nmheim:
accidentally approved, meant to just comment. sorry ;)

pyproject.toml
dominikandreasseitz: changed it back

PR #32, opened 2023-04-12
Add `torch.nn.Module` based API

nmheim:
Sorry for the large diff... except for the new stuff in [`pytorch/modules`](https://github.com/pasqal-io/PyQ/tree/nh/modules/pyqtorch/modules) its just formatting, because I added `pre-commit` to CI as well.

[`examples/bench.py`](https://github.com/pasqal-io/PyQ/tree/nh/modules/examples/bench.py) contains exemplary usage and comparison to the current API.

It seems like we are getting 2x speedup on a single `RX` gate on a 10-qubit state for batch sizes 10-1000 on a GPU.
The same seems to give 10x-100x speedup for batchsize=10000, but not idea why, I still have to verify it. In any case, seems like we are getting faster and we can now do:

```python
circ = QuantumCircuit(...).to(device)
state = zero_state(n, batch_size, device)
thetas = {"phi": torch.rand(batch_size, device)}

circ(thetas, state)
```


Roland-djee (approved):
Looks nice to me :)
Thanks.

PR #34, opened 2023-04-24
pypi install

arthurfaria:


nmheim:
I think we should keep testing different python versions :) closing this one

PR #38, opened 2023-05-01
Fix hamevo mutating the input state

dominikandreasseitz:


pyqtorch/core/batched_operation.py
```diff
@@ -625,6 +626,6 @@ def batched_hamiltonian_evolution_eig(
                 torch.conj(eig_vectors.transpose(0, 1)),
             )
 
-    state = _apply_batch_gate(state, evol_operator, qubits, N_qubits, batch_size_h)
+    _state = _apply_batch_gate(state, evol_operator, qubits, N_qubits, batch_size_h)
```
nmheim, line 629: this looks like the cloned `_state` is not used in the apply batch gate. shouldn't we be doing
```python
return _apply_batch_gate(_state, ...)
```

nmheim (commented):
Nice, thanks a lot! Small comment, otherwise if nobody is relying on this behaviour then I am good with this:)

pyqtorch/core/batched_operation.py
nmheim: but then [this](https://github.com/pasqal-io/PyQ/pull/38/files/cfd5480cff1fd1c995be298b596d59adc6a8b6ba#diff-ab0c36d47e78d3b05edec946b8cf111fed4c8107d78db7561a7c6831a71d275dR543) is not needed

PR #41, opened 2023-05-02
Rework module API

nmheim:


madagra:
Thanks @nmheim. It looks good to me!

PR #42, opened 2023-05-04
Tutorial notebooks

nmheim:


madagra:
@nmheim This is awesome, thanks! The only small thing I would add is to show how to create a custom variational circuit, i.e. show how to define a `nn.Parameter` which then becomes trainable during the optimization.

PR #47, opened 2023-05-10
Add hamevo with torch.linalg.matrix_exp

jpmoutinho:
Current hamevo with diagonalization has unstable gradients in some optimization routines. Using `torch.linalg.matrix_exp` seems to be quite fast and should solve the unstable gradients.

- [x] Add first module
- [x] Include diagonal check not to run matrix_exp on diagonal hamiltonians
- [x] Add more unit tests


jpmoutinho:
Ok this is ready for review whenever you can @dominikandreasseitz. The tests are passing locally, but github is getting an internal error for some reason.

PR #48, opened 2023-05-11
Make `QuantumCircuit.init_state` return state on correct device

nmheim:
the `QuantumCircuit._device` method was always returning `"cpu"` because we don't have parameters in the circuit anymore. now we are checking registered buffers which should work correctly (as long as we are working with a single device).


Roland-djee (approved):
Thanks !

Issue #36, opened 2023-05-01
Overload the * operator of modules.RotationsGate to allow for gate composition


dominikandreasseitz:
**Update for clarity: This issue is about the Module API.**
We want to be able to do arbitrary gate composition for example for the U gate like:

```python
def U(*args: Any, **kwargs:Any) -> QuantumCircuit:
    return RZ( *args, **kwargs) * RY(*args, **kwargs) * RZ(*args, **kwargs)
```

PedroCarreiras:
I ended up changing the version of python I was using. That allowed me to use `pytest` seamlessly, even though I still haven't manage to get hatch to work. Anyways, I think I got something worth sharing to see if this is the intended functionality. I will open a pull request shortly.

nmheim:
closed via #64

Issue #46, opened 2023-05-08
Class hierarchy for gates


nmheim:
It would be great to have a proper interface from which every gate can inherit. Something like

```python
class AbstractGate(torch.nn.Module):
  def __init__(self, gate: str, qubits: ArrayLike, n_qubits: int):
    #...

  @abstractmethod
  def matrices(self, thetas: torch.Tensor) -> torch.Tensor:
    pass

  @abstractmethod
  def apply(self, matrices: torch.Tensor, state: torch.Tensor) -> torch.Tensor:
    pass

  def forward(self, state, thetas):
    return self.apply(self.matrices(thetas), state)
 
  def extra_repr(self):
    # ...


class RotationGate(AbstractGate)
  # ...
```

Goals:
- avoid current code duplication
- force new gates to implement a common interface

PR #64, opened 2023-05-27
Arbitrary Gate Composition

PedroCarreiras:
**Feature**: Allow for the composition of arbitrary gates through the \* operator.

Includes an example notebook.

**Todo**:
- Allow for composition of arbitrary gates
- Enforce DRY code
- Increase number of examples
- Add unit test

**Resolves**: #36


dominikandreasseitz:
can you add a unit test which double-checks your implementation with the current implementation of U-gate? and while youre at it, doublecheck if RX(theta),RY(epsilon),RZ(omega) == U(theta,epsilon,omega) == GateCompUgate(theta,epsilon,omega) ?

PedroCarreiras:
Right, I can see that what was meant to be implemented is not what I had in mind.

To apply RZ(theta)*RY(epsilon)*RZ(omega) == U(omega,epsilon,theta), which is what I understand the correct behavior to be, I'd need to create a class that specifies a \_\_mul\_\_() method. I can't see an easy way to that without breaking things all over the place. And by easy I mean making as little changes as possible to existing code.

I have taken a shot at creating a subclass of torch.Tensor and also at creating a wrapper object. Neither worked.
Ideally I'd have \_apply\_gate() return an object that includes both the state and the gate matrix **if** it were going to call \_\_mul\_\_(), otherwise it would return a torch.Tensor as usual (with the normal \_\_mul\_\_() implementation). And that is a big if as it probably can't be implemented (as far as I know).

Well, to clarify, I am able to get the desired functionality, just not without breaking at least 7 of the existing tests.

Anyways, there is a workaround, which is to use the @ operator instead. It is mapped to the \_\_matmul\_\_() method which is quite handy as the @ operator doesn't seem to be widely used. This means it doesn't break anything to mess with it. However, in needs to be taken into account that using this operator on the returned object from \_apply\_gate() might cause issues.

tl:dr - I created a class named State (there's possibly a better name) that inherits from torch.Tensor which allows the use of the @ operator to concatenate gates. Gates now return an instance of said class but otherwise functionas as normal torch.Tensor.

I'll keep thinking about this, it might require major changes though (and I hope that's just my inexperience talking).
Also, I'll create the unit test soon.

PedroCarreiras:
Hey! That is what I thought at first. The problem is that the functions given as examples above, which take angles as inputs, return the state after applying the respective gate. I am having trouble seeing the connection between the existing RX, RY, RZ, etc classes and the functions also called RX, RY, RZ, etc.

It seems like the intended feature does not match the given example.

nmheim:
Ah, yes, sorry, we are currently moving away from the function versions of RX, RY, RZ, etc. and will use only the classes as the main user interface. We should mention this in the docs. I updated the example with an import statement to indicate that they are the classes defined in pyqtorch.modules, meaning they will return gate objects, not tensors.

sorry for the confusion here!

PedroCarreiras:
Now it was me who didn't explain myself correctly ahah. I meant the confusion came from the following comment:

> can you add a unit test which double-checks your implementation with the current implementation of U-gate? and while youre at it, doublecheck if RX(theta),RY(epsilon),RZ(omega) == U(theta,epsilon,omega) == GateCompUgate(theta,epsilon,omega) ?

These are the RX, RY and RZ functions right? With the classes, if I am not mistaken, I should expect something along the lines of:
```python
from pyqtorch.modules import U, RY, RZ

composed_U = RZ([0], 1)*RY([0], 1)*RZ([0], 1)
composed_U.matrices(theta,epsilon,omega) === U.matrices(theta,epsilon,omega)
```
Returning a circuit (as opposed to a new gate object) is a bit unintuitive to me but I guess it would allow the multiplication of gates applied on different qubits, since the circuit object would "keep track" of everything:

```python
from pyqtorch.modules import RX, RY, QuantumCircuit

RX([0], 2) * RY([1], 2)

> QuantumCircuit(
  (operations): ModuleList(
    (0): RX(qubits=[0], n_qubits=2)
    (1): RY(qubits=[1], n_qubits=2)
  )
)
```
Multiplying two circuits would also be possible and it would chain the operations into one ModuleList. Perhpas using the + operator here would make more sense though, since it is commonly used like this in concatenation of strings and lists.

If this is correct, then even my first implementation wasn't quite right but it was close and I can work on it.

nmheim:
> > can you add a unit test which double-checks your implementation with the current implementation of U-gate? and while youre at it, doublecheck if RX(theta),RY(epsilon),RZ(omega) == U(theta,epsilon,omega) == GateCompUgate(theta,epsilon,omega) ?

yeah, good point that was a bit of a misleading comment... :/

> With the classes, if I am not mistaken, I should expect something along the lines of:
> 
> ```python
> from pyqtorch.modules import U, RY, RZ
> 
> composed_U = RZ([0], 1)*RY([0], 1)*RZ([0], 1)
> composed_U.matrices(theta,epsilon,omega) === U.matrices(theta,epsilon,omega)
> ```

Yes, this would be great, just our `QuantumCircuit` does not have a `matrices` method. If you feel like it you could take a stab at implementing it in the end of course!:)

> Multiplying two circuits would also be possible and it would chain the operations into one ModuleList. Perhpas using the + operator here would make more sense though, since it is commonly used like this in concatenation of strings and lists.

Yes, this is exactly what I had in mind! I see your point about `+`, I think it kind of makes sense though to use `*` here because it resembles the math, so `XY` would be implemented as `X([0],1) * Y([0],1)`. In the future we might actually use the `+` operator for matrix addition as well.

PedroCarreiras:
Hey! I think I got it working. I apologize for the long commit message instead of doing multiple commits. Git was being very weird, deleting all changes and updating some md files without changing anything that I could tell... Oh well.

pyqtorch/modules/circuit.py
```diff
@@ -40,6 +40,22 @@ def __init__(self, n_qubits: int, operations: list):
...
+            raise ValueError(
+                (
+                    f"Number of Qubits don't match. "
+                    f"Left gate is applied on a {self.n_qubits} qubit system and "
+                    f"right gate is applied on a {other.n_qubits} qubit system."
+                )
+            )
+
+        if isinstance(other, QuantumCircuit):
+            return QuantumCircuit(self.n_qubits, other.operations.append(self.operations))
```
nmheim, line 54: Is the order correct here? I would expect that doing `circuit1 * circuit2` would stick the second circuit behind the first. This is the other way round, right?

pyqtorch/modules/circuit.py
```diff
@@ -40,6 +40,22 @@ def __init__(self, n_qubits: int, operations: list):
...
+
+        else:
+            return NotImplemented
```
nmheim, line 57: It would be great to also be able to do `circuit * gate`

pyqtorch/modules/circuit.py
```diff
@@ -56,6 +72,12 @@ def _device(self) -> torch.device:
     def init_state(self, batch_size: int) -> torch.Tensor:
         return zero_state(self.n_qubits, batch_size, device=self._device)
 
+    def matrices(self, thetas: torch.Tensor) -> torch.Tensor:
+        matrix = self.operations[0].matrices(thetas[-1]).squeeze(2)
+        for i, operation in enumerate(self.operations[1:]):
+            matrix = matrix @ operation.matrices(thetas[-2 - i, :]).squeeze(2)
```
nmheim, line 78: This is only correct if all operations act on the same qubit index. I think we could make a separate PR for consistent matrix behavior of pyqtorch. There would not be a unitary hack reward on it but you could still work on it if you like?:)

pyqtorch/modules/parametric.py
```diff
@@ -9,11 +9,11 @@
...
 )
 from pyqtorch.core.utils import OPERATIONS_DICT
+from pyqtorch.modules import QuantumCircuit
+from pyqtorch.modules.utils import rot_matrices
 
 
-class RotationGate(Module):
-    n_params = 1
-
+class Gate(Module):
```
nmheim, line 16: you are kind of addressing #46 here as well;) depending on how much work you want to do here you could give the `AbstractGate` idea a shot to properly formalize this class hierarchy we have in pyq. But then you will likely have to touch almost all of the gates. Primitive/rotation/controlled etc. so I think it should be in a separate PR.

pyqtorch/modules/parametric.py
```diff
@@ -22,6 +22,51 @@ def __init__(self, gate: str, qubits: ArrayLike, n_qubits: int):
         self.register_buffer("imat", OPERATIONS_DICT["I"])
         self.register_buffer("paulimat", OPERATIONS_DICT[gate])
 
+    def __mul__(self, other: RotationGate | QuantumCircuit) -> QuantumCircuit:
```
nmheim, line 25: This should also work for the primitive `X`/`Y`/etc. gates

pyqtorch/modules/parametric.py
```diff
@@ -22,6 +22,51 @@ def __init__(self, gate: str, qubits: ArrayLike, n_qubits: int):
...
+        if isinstance(other, RotationGate):
+            return QuantumCircuit(self.n_qubits, [self, other])
+
+        if isinstance(other, QuantumCircuit):
+            return QuantumCircuit(self.n_qubits, other.operations.insert(0, self))
+
+        else:
+            return NotImplemented
+
+    def __rmul__(self, other: QuantumCircuit) -> QuantumCircuit:
```
nmheim, line 44: I think this one should go into the `QuantumCircuit` itself, see one of the comments above about `circuit*gate`

nmheim (commented):
Nice! This is starting to look good:) additionally to the original issue you have started to address #46 and added the matrices method to the quantum circuit. I think those two things should be addressed in a separate PR, otherwise we will have a lot of changes in this one.

I would suggest to write one (class independent) function that does the multiplication of circuits/gates, which can then be called in the `__mul__` methods. Like that we avoid code duplication and changes in the class hierarchy. Then we could make a separate MR to fix said class hierarchy and add the matrices methods everywhere. If you like you could take a shot at that second PR, what do you think?

and finally, we will need some unit tests for this, could you add some?

PedroCarreiras:
Very well, thanks! All of that makes sense, I'll get on it right away!

pyqtorch/modules/parametric.py
PedroCarreiras: I think I'll start here since it would make annotations much more compact.

pyqtorch/modules/parametric.py
PedroCarreiras: Having an `AbstractGate` class as a parent to all gate classes would help here.

pyqtorch/modules/circuit.py
PedroCarreiras: Indeed. I thought about that but forgot to do anything about it. I'm ok with making a separate PR later to fix that!

pyqtorch/modules/circuit.py
PedroCarreiras: Right. The way I had thought about it was a gate being applied on a circuit (aka after every other gate). But it makes sense to allow a gate to be inserted anywhere, in this case at the beginning. Is this what you meant?

pyqtorch/modules/circuit.py
PedroCarreiras: I used the moduleList in a first-in, first-applied way. However, I realized now that I messed up the matrices method, which works the other way around. Oops.

pyqtorch/modules/circuit.py
nmheim: also thought about this a bit more and added an issue here #71. we will have to refactor the other matrix methods for this as well.

PedroCarreiras:
Hey, sorry for the delay. I think I addressed everything (there were indeed a bunch of misunderstandings from my side).
Is the unit test acceptable?

pyqtorch/modules/circuit.py
nmheim: This method is not correct and should be removed:)

tests/test_modules.py
```diff
@@ -105,3 +105,20 @@ def test_U_gate(batch_size: int) -> None:
...
+            [torch.pi / angle_denominators[0]],
+            [torch.pi / angle_denominators[1]],
+            [torch.pi / angle_denominators[2]],
+        ]
+    )
+
+    customUGate = circ.matrices(thetas)
+    UGate = pyq.U(qubits=[0], n_qubits=1).matrices(thetas)
+
+    assert customUGate.allclose(UGate)
```
nmheim, line 124: This test is great, but given that we have to remove `QuantumCircuit.matrices` it would be easier to test that the states match here.

Additionally, I will write some tests in pseudocode below that should be added in a second `test_some_test_name` function.
```python
assert X([0],1) * Y([0],1) == QuantumCircuit(1, [X(..), Y(..)])

# we have to deal with different number of incoming qubits
# i.e. output circuit should have the larger number of qubits of the two
assert X([1],2) * Y([0],1) == QuantumCircuit(2, [X(..), Y(..)])

# test the same thing for RX/RY

assert QuantumCircuit(1, ...) * RX([0],1) == QuantumCircuit(1, [..]+[X(..)])

# also test that circuit is getting more qubits
assert QuantumCircuit(1, ...) * RX([1],2) == QuantumCircuit(2, ...)

# probably throw and error in the case of `gate * circuit`?
X([1],2) * QuantumCircuit(1, ...) == QuantumCircuit(2, ...)

# `QuantumCircuit`s should be composable
assert QuantumCircuit(1, ...) * QuantumCircuit(2, ...) == QuantumCircuit(2, ...)

# and finally, just as a sanity check, run every `QuantumCircuit` resulting from a composition
# to make sure it works. does not have to be compared to a numerical result,
# this is just to check that is runs and does not error
```

nmheim (changes requested):
Thanks for the updates! As soon was the unit tests I suggested are running I think this should be ready to be merged :)

nmheim (approved):
I think this is looking pretty good now! The `is_same_circuit`/`is_same_gate` methods should be solved a little bit different for the future, but given that this is the end of the unitaryhack and that you already got another MR merged (#46) we can merge this one for now and I'll open an issue for minor thing of the equality checks. so you'll get your bounty! thanks again for all the work @PedroCarreiras

PR #69, opened 2023-06-01
Unified the three Hamiltonian Evolution Classes

### LINKED ISSUE #53: Unify the three hamiltonian evolution classes into one `HamiltonianEvolution`
**Update for clarity: This issue is about the Module API.**
There are currently 3 different ways to implement the Hamiltonian evolution in PyQ. We should allow the user to select the type of operation when initializing the gate operation in the quantum circuit constructor instead of doing it through three different classes.

Unify the three classes in `pyqtorch.modules.hamevo` into one which accepts an Enum that determines which method is used.

```python
class HamiltonianEvolution(Module):
    def __init__(
        self,
        qubits: Any,
        n_qubits: int,
        n_steps: int = 100,
        # the type should be an enum that also accepts string inputs
        type: HamEvoType=HamEvoType.RK4
    ):
```

zain2864:
Upgrade: Instead of creating instances of HamEvo, HamEvoEig, or HamEvoExp, users can create an instance of HamiltonianEvolution and specify the desired Hamiltonian evolution method with the type parameter.  The class accepts an Enum (HamEvoType) during initialization that determines the Hamiltonian evolution method to be used. The HamEvoType Enum includes the values RK4, EIG, and EXP, each corresponding to one of the existing Hamiltonian evolution classes. The class also accepts string inputs for the Hamiltonian evolution method type and converts them into the corresponding Enum values.

fixes #53


pyqtorch/modules/hamevo.py
```diff
@@ -165,47 +166,23 @@ def apply(self, state: torch.Tensor) -> torch.Tensor:
...
         return _apply_batch_gate(state, evol_operator, self.qubits, self.n_qubits, batch_size_h)
 
+class HamEvoType(Enum):
+    RK4 = HamEvo
+    EIG = HamEvoEig
+    EXP = HamEvoExp
 
 class HamiltonianEvolution(Module):
-    def __init__(self, qubits: Any, n_qubits: int, n_steps: int = 100):
+    def __init__(self, qubits: Any, n_qubits: int, n_steps: int = 100, type: HamEvoType = HamEvoType.RK4):
```
madagra, line 175: The type of the `type` argument should be `HamEvoType | str` to be consistent. Furthermore, it is better to avoid naming the variable `type` since it clashes with the Python standard library function. Maybe `hamevo_type`?

pyqtorch/modules/hamevo.py
```diff
@@ -165,47 +166,23 @@ def apply(self, state: torch.Tensor) -> torch.Tensor:
...
         super().__init__()
         self.qubits = qubits
         self.n_qubits = n_qubits
         self.n_steps = n_steps
+        if isinstance(type, str):
+            type = HamEvoType[type.upper()]
```
madagra, line 181: You could add a `try...except` block here to catch the case where the user passes an invalid string and send a clear error message.

madagra (commented):
Thanks @zain2864 for this PR. Before moving forward, could you please add some unit tests?

dominikandreasseitz:
Hi, great work done here! Thanks! Could we maybe get a docstring similar to MR https://github.com/pasqal-io/PyQ/pull/72 for HamEvo? just a short description of those 3 options

zain2864:
> 

Thank you so much! Should add I documentation in the hamevo.py file in pyqtorch/modules and/or create a new file in docs \documentation\modules. However, since MR #72 has not been merged yet, I think I should wait if it is the latter.

dominikandreasseitz:
a docstring + a hamevo.md file would suffice imo! thanks!

docs/hamevo.md
```diff
@@ -0,0 +1,52 @@
...
+```python
+# Instantiate HamiltonianEvolution with RK4 string input
+hamiltonian_evolution = HamiltonianEvolution(qubits, n_qubits, 100, "RK4")
+# Use the HamiltonianEvolution instance to evolve the state
+output_state = hamiltonian_evolution(H, t, state)
+
+# Instantiate HamiltonianEvolution with HamEvoType input
+H_evol = HamiltonianEvolution(qubits, n_qubits, 100, HamEvoType.EIG)
+# Use the HamiltonianEvolution instance to evolve the state
+output_state = H_evol(H, t, state)
```
madagra, line 52: @zain2864 Can you make this example self-contained by adding relevant imports and the definition of the initial state?

madagra (commented):
@zain2864 Almost ready to be merged. Just a few minor comments and making sure the CI pipeline is passing.

docs/hamevo.md
zain2864: Yes! I have split it into two examples, one for showing string input and the other for showing the enum input. Both contain relevant imports and the definitions of the initial state and evolution time tensor. Let me know if there are any other additions or changes needed in the hamevo.md file

madagra: Thank you, @zain2864. It looks good to me now.

pyqtorch/modules/hamevo.py
```diff
@@ -166,46 +242,101 @@ def apply(self, state: torch.Tensor) -> torch.Tensor:
...
+        >>>output = H_evol(H, t, state) # Use the HamiltonianEvolution instance to evolve the state
+
+    """
+
+    def __init__(
+        self,
+        qubits: Any,
+        n_qubits: int,
+        n_steps: int = 100,
+        hamevo_type: Union[HamEvoType, str] = HamEvoType.RK4,
```
madagra, line 294: @zain2864 Last minor comment: please set the default to `HamEvoType.HamEvoExp` since it is the most reliable method.

madagra (commented):
@zain2864 Make sure that the CI is passing (currently the linting stage does not, see the contribution guidelines on how to apply linting). After that, it looks ready for merging.

madagra (approved):

PR #70, opened 2023-06-01
Class hierarchy for gates

PedroCarreiras:
**Refactor **: Restructure gates to inherit from an AbstractGate class to reduce code duplication and allow for easier modifications.

Resolves: #46


pyqtorch/modules/parametric.py
```diff
@@ -38,28 +34,10 @@ def forward(self, state: torch.Tensor, thetas: torch.Tensor) -> torch.Tensor:
         return self.apply(mats, state)
 
     def extra_repr(self) -> str:
-        return f"qubits={self.qubits}, n_qubits={self.n_qubits}"
+        return super().extra_repr()
```
nmheim, line 37: All of those methods can be removed completely now, because you are only calling the inherited function from `AbstractGate` anyway:)

pyqtorch/modules/utils.py
```diff
@@ -1,6 +1,34 @@
 from __future__ import annotations
 
+from abc import ABC, abstractmethod
+
 import torch
+from numpy.typing import ArrayLike
+from torch.nn import Module
+
+
+class AbstractGate(ABC, Module):
```
nmheim, line 10: Let’s move this to a file `abstract.py`

nmheim (commented):
Great work! Thanks a lot for opening this one as well! Just some polishing and tests missing:)

just to make sure we are all on the same page wrt unitary hack: this PR is not part of the unitary hack anymore so you won’t get a monetary reward for this. But your contributions are of course still super welcome and we are very happy about them:)

pyqtorch/modules/parametric.py
PedroCarreiras: Right, I initially made it an abstract method until I realized they all had the same method. I forgot to remove all of those afterward.

pyqtorch/modules/utils.py
PedroCarreiras: On it!

nmheim:
Thanks a lot for this one @PedroCarreiras ! :)

PR #72, opened 2023-06-04
doc strings for most important gates/circuits #54

prilcool:
Added new doc strings for important gates & functions
-Added documentation nav in mkdocs

PR #73, opened 2023-06-07
Fix certain gates not return double-precision results

jpmoutinho:


pyqtorch/modules/parametric.py
```diff
@@ -81,9 +81,9 @@ def __init__(self, qubits: ArrayLike, n_qubits: int):
 
     def matrices(self, thetas: torch.Tensor) -> torch.Tensor:
         if thetas.ndim == 1:
-            thetas = thetas.unsqueeze(1)
+            thetas = thetas.unsqueeze(1).cdouble()
         assert thetas.size(0) == 3
-        phi, theta, omega = thetas[0, :], thetas[1, :], thetas[2, :]
+        phi, theta, omega = thetas[0, :].cdouble(), thetas[1, :].cdouble(), thetas[2, :].cdouble()
```
dominikandreasseitz, line 86: hmm, a bit ugly to always have to call this, isnt it? can we maybe find a better way. maybe there is something in the torch docu which enables the required precision. to me it seems like that it should maybe be handled on the user-side which passes the thetas?

dominikandreasseitz:
thank you @jpmoutinho , i added a comment, maybe we can find a different way to deal with this instead of always having to call .cdouble(). tagging @madagra here, maybe he has an idea

pyqtorch/modules/parametric.py
jpmoutinho: I agree it's ugly yes. We could try to do `torch.set_default_dtype(torch.float64) like documented here: https://pytorch.org/docs/stable/generated/torch.set_default_dtype.html.

I guess in theory it might allows us to remove all the `cdouble` both from the thetas and maybe also from the `dtype = cdouble` when initializing torch tensors, but I assume this would have to be added to every single file in the library.

For the theta's here the reason I don't think it should be handled by the user is that for any of these parametric gates if the user just sets "angle = 1.0" or something with python float, which seems very natural, it is currently cast to 32 bit precision. Even just one gate that does this will ruin the precision of the whole circuit.

jpmoutinho: Alright I've tested it now and `torch.set_default_dtype(torch.float64)` seems to do the trick! I removed the `.cdouble()` I added (but not the ones from the torch tensor initializations).

Would it make sense to have this line added by default when linting, or you think that could be dangerous in some way?

dominikandreasseitz: yeah we should test it on GPU for sure

pyqtorch/core/utils.py
```diff
@@ -20,6 +20,8 @@
 import torch
 from numpy.typing import NDArray
 
+torch.set_default_dtype(torch.float64)
```
dominikandreasseitz, line 23: maybe it would be enough to set it in circuit.py?  it would enforce it an all submodules used in the file as far as i understand?

jpmoutinho: Let me test :)

jpmoutinho: Doesn't seem to work unfortunately. I believe you need the line in every file where a `torch.tensor()` is initialized.

dominikandreasseitz: ok thank you for testing this :) hm, so we just put it on top of every file (which wouldnt be a lot), and then also externally passed thetas will automatically converted to this?

jpmoutinho: Well for now we can just keep it in these two files, which seems to cover all the gates being created. But yes, this fixes the thetas.

If we always include it by default in a new file we could in theory be more relaxed with the `dtype=torch.cdouble` used everywhere, as torch should initialize everything with `float64` and automatically convert to `complex128` when a complex number is detected, but I did not test this. @madagra also mentioned we could at some point have this be a global setting we could control, as we could then choose to run a given simulation in `float32 / complex64` to save memory if needed, but this would need to be properly tested in a new MR.

madagra: Wouldn't be better to put it in the `__init__.py` of the repo? In this way, we are sure it is always called no matter what the user does. And yes, having it as a sort of tweakable configuration would be nice (see for instance [this library](https://github.com/lululxvi/deepxde/blob/master/deepxde/config.py) for a possible implementation: ). For now, let's leave it like this, though.

madagra (approved):
@jpmoutinho I have a small comment on the placement of the `set_default_type` call but overall this is ready for merging.

pyqtorch/core/utils.py
jpmoutinho: I tested now to move it to the `__init__.py` of the repo but it's not producing the same thing, the results still come out as single precision.

nmheim:
I am a bit puzzled by why this is necessary at all, because it should be taken care of by `model.to` and firends, but I've not been following this during holiday, so I am ok if you want to merge this. just out of curiosity... which gates were not returning double precision by default before?

nmheim:
> @nmheim the gates were:
> 
>     * the Hadamard gate, caused by `torch.sqrt(torch.tensor(2))` in `core/utils.py`
> 
>     * the T gate, caused by `torch.exp(torch.tensor(1j) * torch.pi / 4)` in `core/utils.py`
> 
>     * the parametric gates in `modules/parametric.py` caused by the handling of the `theta` parameter, which when passed by a user as e.g. a generic python float would get converted to `torch.float32`

If I try the Hadamard gate I am getting single/double precision as I would expect. I think I am missing something 😅 
```python
In [15]: import pyqtorch.modules as pyq

In [16]: g = pyq.H([0],1)

In [18]: list(g.named_buffers())
Out[18]: 
[('matrix',
  tensor([[ 0.7071+0.j,  0.7071+0.j],
          [ 0.7071+0.j, -0.7071+0.j]], dtype=torch.complex128))]

In [19]: g(pyq.zero_state(1))
Out[19]: 
tensor([[0.7071+0.j],
        [0.7071+0.j]], dtype=torch.complex128)
```
those all have double precision and if I move the gate to single precision I am also getting expected results
```python
In [25]: f = g.to(dtype=torch.cfloat)

In [26]: f(pyq.zero_state(1, dtype=torch.cfloat)).dtype
Out[26]: torch.complex64
```

I don't want to not merge this, don't get me wrong, I'd just like to understand where the error was :P

jpmoutinho:
Of course @nmheim, I should have added an example script in the first place :). You are right things are being cast to the correct precision type, but some of the intermediate calculations were being done with 32-bit precision, causing the error to propagate to the final result either way. You can try the following script:

```
import torch
import pyqtorch.modules as pyq

def norm(wf):
    return abs(sum((wf.conj() * wf).squeeze())) ** 2

g = pyq.H([0], 1)

exact_norm = torch.tensor(1.0, dtype = torch.float64)

wf_init = pyq.zero_state(1)

wf = g(pyq.zero_state(1))

print(abs(exact_norm - norm(wf)))
```

On `main` it should return an error in the wavefunction normalization of `E-08` while on this branch it should be `E-16`.

pyqtorch/core/utils.py
madagra: OK, interesting. Did you put in the `__init__.py` of the full repo, right? I mean `pyqtorch/__init__.py` and not the one of a submodule. In any case, if it does not work, let's keep it as it is and merge this one.

madagra (approved):
I think this is ready to merge.

dominikandreasseitz:
this MR is still open? :D

pyqtorch/core/utils.py
jpmoutinho: Yes @madagra I tried in `pyqtorch/__init__.py`

madagra:
> this MR is still open? :D

It should be merged, indeed, @jpmoutinho

jpmoutinho:
@dominikandreasseitz do you want to deal with the version / release here? Or leave it for another MR?

madagra:
> @dominikandreasseitz do you want to deal with the version / release here? Or leave it for another MR?

@jpmoutinho @dominikandreasseitz I think we can wait for #72 before bumping the version and do a new release?

Issue #54, opened 2023-05-14
Write doc strings for most important gates/circuits


dominikandreasseitz:
**Update for clarity: This issue is about the Module API.**
The most important functions and classes in `pyqtorch.modules` should get a doc string.
A doc string should look like below

```python
def zero_state(
    n_qubits: int,
    batch_size: int = 1,
    device: str | torch.device = "cpu",
    dtype: torch.dtype = torch.cdouble,
) -> torch.Tensor:
    """Describe what function does.
    
    Arguments:
        n_qubits: __description__
        ...

    Returns:
        torch.Tensor: __description__

    Examples:
    ```python exec="on" source="above" result="json"
    from import pyqtorch.modules as pyq

    state = pyq.zero_state(n_qubits=2)
    print(str(state)) # markdown-exec: hide
    ```
    """
```

Where optionally we could include examples that are run during the docs build process (see `exec=on` arguments inside the code block) which would ensure that the examples work and can be directly copy pasted by users to play with them. In order to enable these runable examples we have to add the [`markdown-exec`](https://pypi.org/project/markdown-exec/) plugin to the `mkdocs.yml`.

prilcool:
very nice  codebase,  I'll  give this a shot and start  working on this now  :)

dominikandreasseitz:
hey, thanks a lot and welcome! have fun solving the issue :)

prilcool:
hey  @dominikandreasseitz  ,
 just a query , do you want the docstrings & example to   appear  before the function  or  inside the function  or after the function

prilcool:
Hi @dominikandreasseitz 

Just checking  , For  modules ,documentation that is generated by mkdocs from the docstring from the docs folder ,  Is  this expected output ?  https://drive.google.com/file/d/11N4TCD4wIvp_E56DcD3PpIznuUl_d6qV/view?usp=sharing

Also when I run  pytest or   hatch -e docs run mkdocs build --clean --strict   it all shows success , 
I am not sure  if it's supposed to show more  additional info or not eg print the output of running the example  due to  this line  under Examples

  `  python exec="on" source="above" result="json`"


Thanks

dominikandreasseitz:
hi @prilcool , thats looks great! submit a PR once you are satisfied so i can build the documentation myself too and hav a look at it. then we can iteratively improve it :) thanks!

PR #74, opened 2023-06-07
docstring added (issue #54)

MihirsinhChauhan:


mkdocs.yml
```diff
@@ -13,6 +13,12 @@ nav:
   - Tutorials:
       - Fitting a function: fit_function.ipynb
       - QAOA : QAOA.ipynb
+  - Documentation:
```
dominikandreasseitz, line 16: can we maybe have this under "Overview".

dominikandreasseitz:
thank you! but the issue has been addressed in https://github.com/pasqal-io/PyQ/pull/72. i will definitely have a look at both and see if we can combine stuff.

MihirsinhChauhan:
should need to do any changes and if it is, please give suggestion regarding same

dominikandreasseitz:
hi @MihirsinhChauhan, unfortunately we cant go ahead with this MR since its a duplicate! thank you so much for wanting to contribute though!

PR #76, opened 2023-06-12
Move `is_same_XXX` methods to `__eq__`

### LINKED ISSUE #75: Equality of circuits/gates
In #64 we introduced a `is_same_circuit`/`is_same_gate` method which sometimes only checks if the types of the circuits are equal, not the values themselves. This should be fixed and moved to `__eq__`.

nmheim:
Closes #75

### DIFF
--- a/pyqtorch/modules/abstract.py
+++ b/pyqtorch/modules/abstract.py
@@ -1,12 +1,13 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
+from typing import Any
 
 import torch
 from numpy.typing import ArrayLike
 from torch.nn import Module
 
-import pyqtorch.modules
+import pyqtorch.modules as pyq
 
 
 class AbstractGate(ABC, Module):
@@ -15,21 +16,20 @@ class AbstractGate(ABC, Module):
         self.qubits = qubits
         self.n_qubits = n_qubits
 
-    def __mul__(
-        self, other: AbstractGate | pyqtorch.modules.QuantumCircuit
-    ) -> pyqtorch.modules.QuantumCircuit:
+    def __mul__(self, other: AbstractGate | pyq.QuantumCircuit) -> pyq.QuantumCircuit:
         if isinstance(other, AbstractGate):
-            return pyqtorch.modules.QuantumCircuit(
-                max(self.n_qubits, other.n_qubits), [self, other]
-            )
-
-        if isinstance(other, pyqtorch.modules.QuantumCircuit):
-            return pyqtorch.modules.QuantumCircuit(
-                max(self.n_qubits, other.n_qubits), [self, *other.operations]
-            )
-
+            return pyq.QuantumCircuit(max(self.n_qubits, other.n_qubits), [self, other])
+        elif isinstance(other, pyq.QuantumCircuit):
+            # using * here because list concatenation is not implemented for ModuleList
+            return pyq.QuantumCircuit(max(self.n_qubits, other.n_qubits), [self, *other.operations])
         else:
-            return NotImplemented
+            return ValueError(f"Cannot compose {type(self)} with {type(other)}")
+
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, type(self)):
+            return False
+
+        return self.qubits == other.qubits and self.n_qubits == other.n_qubits
 
     @abstractmethod
     def matrices(self, tensors: torch.Tensor) -> torch.Tensor:
@@ -47,10 +47,3 @@ class AbstractGate(ABC, Module):
 
     def extra_repr(self) -> str:
         return f"qubits={self.qubits}, n_qubits={self.n_qubits}"
-
-    def is_same_gate(self, other: AbstractGate) -> bool:
-        return (
-            self.qubits == other.qubits
-            and self.n_qubits == other.n_qubits
-            and type(self) == type(other)
-        )
--- a/pyqtorch/modules/circuit.py
+++ b/pyqtorch/modules/circuit.py
@@ -43,15 +43,25 @@ class QuantumCircuit(Module):
 
     def __mul__(self, other: AbstractGate | QuantumCircuit) -> QuantumCircuit:
         if isinstance(other, QuantumCircuit):
-            return QuantumCircuit(
-                max(self.n_qubits, other.n_qubits), self.operations.extend(other.operations)
-            )
+            n_qubits = max(self.n_qubits, other.n_qubits)
+            return QuantumCircuit(n_qubits, self.operations.extend(other.operations))
 
         if isinstance(other, AbstractGate):
-            return QuantumCircuit(max(self.n_qubits, other.n_qubits), self.operations.append(other))
+            n_qubits = max(self.n_qubits, other.n_qubits)
+            return QuantumCircuit(n_qubits, self.operations.append(other))
 
         else:
-            return NotImplemented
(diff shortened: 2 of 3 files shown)


pyqtorch/modules/abstract.py
```diff
@@ -15,21 +16,24 @@ def __init__(self, qubits: ArrayLike, n_qubits: int):
...
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, type(self)):
+            return False
 
-        else:
-            return NotImplemented
+        return self.qubits == other.qubits and self.n_qubits == other.n_qubits
+
+    def __hash__(self):
+        return super().__hash__()
```
nmheim, line 36: if we don't add this we are getting errors like
```
TypeError: unhashable type: 'X'
```
I am not sure why they are popping up when adding `__eq__` methods. I have no idea how pytorch is implementing this (also could not find out in a reasonable amount of time), but our gates/circuits should strictly speaking not even have been hashable in the first place, because they have the `qubits` field which is a `list` (which is not hashalbe)...

pyqtorch/modules/circuit.py
```diff
@@ -43,15 +43,26 @@ def __init__(self, n_qubits: int, operations: list):
...
+    def __key(self) -> tuple:
+        return (self.n_qubits, *self.operations)
+
+    def __eq__(self, other: Any) -> bool:
+        if isinstance(other, QuantumCircuit):
+            return self.__key() == other.__key()
+        return NotImplemented
+
+    def __hash__(self) -> int:
+        return hash(self.__key())
```
nmheim, line 65: the only problem with this in our case is that changing the `operations: ModuleList` is going to change the hash which is not what python/pytorch is expecting... pytorch is returning a different hash for every created object, so also different for `ModuleList`s that contain the same elements. we could go for `hash() = super().__hash__()` or something, but then `__eq__` and `__hash__` have inconsistent behaviour... not sure which way to go

dominikandreasseitz:
thanks for this @nmheim

pyqtorch/modules/circuit.py
dominikandreasseitz: i would rather go for just hashing the names of the gates in the module list. this should return true if the circuit is the same right? not necessarily if the (random) param values are the same

nmheim: I am not sure if I understand what you mean by names of the gates... for e.g. `X([0],1)` just hash `X`? then `X([1],2)` would have the same hash...

nmheim:
This is ready to be merged IMO :)

madagra:
@nmheim What is preventing from merging this one?

madagra:
> This is ready to be merged IMO :)

Yes, I would go ahead.

pyqtorch/modules/abstract.py
```diff
@@ -15,21 +16,26 @@ def __init__(self, qubits: ArrayLike, n_qubits: int):
         self.qubits = qubits
         self.n_qubits = n_qubits
 
-    def __mul__(
-        self, other: AbstractGate | pyqtorch.modules.QuantumCircuit
-    ) -> pyqtorch.modules.QuantumCircuit:
+    def __mul__(self, other: AbstractGate | pyq.QuantumCircuit) -> pyq.QuantumCircuit:
```
Roland-djee, line 19: Is there any other `QuantumCircuit` that clashes with this one ? If not, I would import it directly since it is obvious that we are dealing with PyQ types.

pyqtorch/modules/abstract.py
```diff
@@ -15,21 +16,26 @@ def __init__(self, qubits: ArrayLike, n_qubits: int):
...
-            return pyqtorch.modules.QuantumCircuit(
-                max(self.n_qubits, other.n_qubits), [self, other]
-            )
+            ml = torch.nn.ModuleList([self, other])
+            return pyq.QuantumCircuit(max(self.n_qubits, other.n_qubits), ml)
+        elif isinstance(other, pyq.QuantumCircuit):
+            ml = torch.nn.ModuleList([self]) + other.operations
+            return pyq.QuantumCircuit(max(self.n_qubits, other.n_qubits), ml)
+        else:
+            return ValueError(f"Cannot compose {type(self)} with {type(other)}")
```
Roland-djee, line 27: `TypeError` ?

pyqtorch/modules/abstract.py
```diff
@@ -15,21 +16,26 @@ def __init__(self, qubits: ArrayLike, n_qubits: int):
...
 
-        if isinstance(other, pyqtorch.modules.QuantumCircuit):
-            return pyqtorch.modules.QuantumCircuit(
-                max(self.n_qubits, other.n_qubits), [self, *other.operations]
-            )
+    def __key(self) -> tuple:
```
Roland-djee, line 29: `__key__` ?

pyqtorch/modules/abstract.py
```diff
@@ -15,21 +16,26 @@ def __init__(self, qubits: ArrayLike, n_qubits: int):
...
+        return (self.n_qubits, *self.qubits)
 
-        else:
-            return NotImplemented
+    def __eq__(self, other: Any) -> bool:
+        if isinstance(other, type(self)):
+            return self.__key() == other.__key()
+        return NotImplemented
```
Roland-djee, line 35: Maybe a message ?

pyqtorch/modules/circuit.py
```diff
@@ -43,15 +43,26 @@ def __init__(self, n_qubits: int, operations: list):
...
 
     def forward(self, state: torch.Tensor, thetas: torch.Tensor = None) -> torch.Tensor:
```
Roland-djee, line 67: `from torch import Tensor` ?

pyqtorch/modules/abstract.py
```diff
@@ -15,21 +16,26 @@ def __init__(self, qubits: ArrayLike, n_qubits: int):
...
-            return pyqtorch.modules.QuantumCircuit(
-                max(self.n_qubits, other.n_qubits), [self, *other.operations]
-            )
+    def __key(self) -> tuple:
+        return (self.n_qubits, *self.qubits)
 
-        else:
-            return NotImplemented
+    def __eq__(self, other: Any) -> bool:
+        if isinstance(other, type(self)):
```
Roland-djee, line 33: `AbstractGate` ? Unless this equality checking works for subtypes ?

Roland-djee (commented):
Hey @nmheim , few comments from my side. Hope this is helpful.

pyqtorch/modules/abstract.py
nmheim: I tried, but getting circular imports for some reason

pyqtorch/modules/abstract.py
nmheim: dunder methods should only be used for things like `__hash__` used by functions such as `hash`. `__key__` is currently not used by any of those, but future python versions could decided to do that

pyqtorch/modules/abstract.py
nmheim: `NotImplemented` is not an error but some other thing that forwards the error. so if the quality check above fails python will try to use other equality methods and throw an appropriate error

pyqtorch/modules/abstract.py
nmheim: This would result in `X==Y` returning true

pyqtorch/modules/abstract.py
nmheim: done

PR #79, opened 2023-06-22
Support py3.11

dominikandreasseitz:


dominikandreasseitz:
youre right, i can re-add it

PR #81, opened 2023-06-22
Correct tensor type

Roland-djee:
Another PR in the same vein: specifying `dtype` to avoid errors:

```python
>       mat_indices[len(qubits) : 2 * len(qubits)] = state_indices[qubits]
E
```

because:

```
qubits = array([0], dtype=object)
```

### DIFF
--- a/pyqtorch/core/utils.py
+++ b/pyqtorch/core/utils.py
@@ -152,7 +152,7 @@ def _apply_batch_gate(
     Same shape as `input_state`
     """
     mat = mat.reshape([2] * len(qubits) * 2 + [batch_size])
-    qubits = N_qubits - 1 - np.array(qubits)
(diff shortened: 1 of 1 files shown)


pyqtorch/core/utils.py
```diff
@@ -150,7 +152,7 @@ def _apply_batch_gate(
     Same shape as `input_state`
     """
     mat = mat.reshape([2] * len(qubits) * 2 + [batch_size])
-    qubits = N_qubits - 1 - np.array(qubits)
+    qubits = np.array(N_qubits - 1 - np.array(qubits), dtype=np.int64)
```
nmheim, line 155: why numpy array here?

Roland-djee: Well, to specify the `dtype`. Without it, it defaults the `dtype` to objects as it can't infer it from `qubits`.

dominikandreasseitz (approved):

PR #82, opened 2023-07-03
add CSWAP gate support

Doomsk:


dominikandreasseitz:
great! thanks for this addition @Doomsk

tests/test_operations.py
```diff
@@ -137,3 +145,28 @@ def test_CRY_state10_controlqubit_0() -> None:
 def test_CRY_state01_controlqubit_0() -> None:
     result: torch.Tensor = operation.CRY(pi, state_01, (1, 0), 2)
     assert torch.allclose(state_11, result)
+
+
+def test_CSWAP_state000_controlqubit_0() -> None:
+    result: torch.Tensor = operation.CSWAP(state_000, (0, 1, 2), 3)
+    assert torch.allclose(state_000, result)
```
nmheim, line 152: the functional API of pyqtorch will be deprecated soon. can we add those tests to `tests/test_modules.py` rather than `test_operations.py`?

dominikandreasseitz: oh yes, thanks @nmheim

awennersteen: Maybe @nmheim (in a a separate PR) you can mark the functional API as deprecated? see e.g. Pulser for how to do that, or pasqal-cloud

tests/test_modules.py
```diff
@@ -189,3 +195,28 @@ def test_circuit_composition(n_qubits: int) -> None:
     )
     truth = pyq.QuantumCircuit(n_qubits, [pyq.X([0], 1), pyq.Y([0], 1), rx, pyq.RY([0], 1)])
     assert circ == truth
+
+
+def test_CSWAP_state000_controlqubit_0() -> None:
+    result: torch.Tensor = func_pyq.operation.CSWAP(state_000, (0, 1, 2), 3)
```
dominikandreasseitz, line 201: @Doomsk youre not using the pyq.modules.CSWAP?

Doomsk: @dominikandreasseitz I'm just following the previous tests, using `operation.<gate>` as `Tensor` to compare the initial and final states, for example, as in `test_operations.py` line 145:

```python
def test_CRY_state01_controlqubit_0() -> None:
    result: torch.Tensor = operation.CRY(pi, state_01, (1, 0), 2)
    assert torch.allclose(state_11, result)
```

Nonetheless, they should produce the same result, since both `operation` and `modules` call from the same source on `OPERATIONS_DICT` at `utils.py`.

dominikandreasseitz: yes but all the other test in test_modules use the module API? would be really nice if you can change it

Doomsk: @nmheim mentioned  the API was going to be deprecated, or was it another API?

dominikandreasseitz: everything in core (except utils) will be deprecated, yes. but i am asking you to use the pyqtorch/modules/ (module API), and your tests use core/operation stuff which will be deprecated

PR #83, opened 2023-07-04
Batching of `HamEvo`s in time

nmheim:


nmheim:
There are surely nicer ways of fixing this... but I kind of need it now for an upstream package as soon as possible

tests/test_module_hamevo.py
jpmoutinho, line 1: Rogue print in line 129.

tests/test_module_hamevo.py
```diff
@@ -167,19 +169,56 @@ def test_hamevo_consistency(get_hamiltonians: Callable) -> None:
...
+            2,
+        ),
+        (  # batchsize 2 | 2
+            Hamiltonian(2),
+            torch.tensor([torch.pi / 4, torch.pi]),
+            torch.tensor([0.5, 1.0]),
+            2,
+        ),
+    ],
+)
```
jpmoutinho, line 200: I don't really like the usage of `Hamiltonian` here for this test just because it's a very strict generator (being diagonal) which could potentially mask incorrect results for other more general cases, but of course I understand why you used it here to compare with the analytical results for specific times. Also since the other tests are also passing with the more general hamiltonian, even if they don't use time batching, I'm ok with leaving it as is.

I feel like the HamEvo code could use a little cleaning anyway and I would be happy to take care of that at some point 😅.

nmheim: yeah, I very much agree. I thought we were going to unify the three evolution methods with #53, but that just resulted in adding another class. I'll write an issue for this and also mention to write better tests

nmheim:
> Did you run any performance tests comparing the `HamEvoExp` with `HamEvoEig` for batched-times? I can also run a few if you wish. Ideally you should see big improvements with `HamEvoEig` if you use just one moderately sized hamiltonian and a large batch of time values.

Not properly. I actually ended up with a slightly smaller runtime for a batch of 1000 and two qubits for `HamEvoExp`, but maybe that was not a good example. this should be investigated at some point 😅

PR #87, opened 2023-07-12
[Feature] Add batched overlap, add batched haar measure random_state, add batched normalization

dominikandreasseitz:


tests/test_modules.py
```diff
@@ -225,3 +225,12 @@ def test_CSWAP_state110_controlqubit_0() -> None:
...
     assert torch.allclose(cswap(state_110), state_101)
+
+
+@pytest.mark.parametrize("n_qubits", [1, 2, 4, 6])
+@pytest.mark.parametrize("batch_size", [1, 2, 4, 6])
+def test_overlap_nqubits(n_qubits: int, batch_size: int) -> None:
+    assert torch.allclose(
+        pyq.overlap(pyq.zero_state(n_qubits, batch_size), pyq.zero_state(n_qubits, batch_size)),
+        torch.ones(batch_size),
+    )
```
madagra, line 236: Can you use some random state instead of the `zero_state` for the test?

madagra (approved):
A part from the small comment looks good to me.

tests/test_modules.py
nmheim: +1

nmheim: why does it only work for `batch_size=1`? the `random_state` looks like its supporting batching

dominikandreasseitz: the overlap.

dominikandreasseitz: fixed it now, overlap works with batched, added loads of tests too for normalizaion checks etc. all batched too

PR #96, opened 2023-08-09
Rename to pyqtorch, Introduce vmap_gate

dominikandreasseitz:


nmheim:
Ohh nice, is this faster than the batch gate we have currently?

tests/test_speed.py
```diff
@@ -0,0 +1,60 @@
...
+    return wrapper
+
+
+@no_type_check
+@timeit
+def time_fn():
+    speed_test_module_circuit()
+
+
+if __name__ == "__main__":
```
dominikandreasseitz, line 59: @nmheim run this file on main and then here, this is certainly not fully correct yet but maybe we are on to something

pyqtorch/modules/circuit.py
```diff
@@ -85,8 +85,11 @@ def forward(self, state: torch.Tensor, thetas: torch.Tensor = None) -> torch.Ten
             torch.Tensor: The output quantum state tensor after applying the circuit operations.
 
         """
-        for op in self.operations:
+
+        def fn(op: torch.nn.Module) -> torch.Tensor:
             state = op(state, thetas)
+
+        map(fn, self.operations)
```
nmheim, line 92: `functools.reduce` would be the way to here I think, but won’t make a difference to the previous for loop I think?

nmheim (commented):
I like where this is going!

tests/test_modules.py
```diff
@@ -95,7 +95,7 @@ def test_controlled_parametrized_gates(batch_size: int, n_qubits: int, gate: str
 
     assert torch.allclose(func_out, mod_out)
 
-
+@pytest.mark.skip
```
awennersteen, line 98: ?

pyqtorch/core/utils.py
```diff
@@ -187,3 +187,32 @@ def _apply_batch_gate(
     state = torch.einsum(einsum_indices, mat, state)
 
     return state
+
+
+def _apply_map_gate(
```
awennersteen, line 192: nice, very interesting proposal!

tests/test_speed.py
```diff
@@ -0,0 +1,60 @@
+from __future__ import annotations
```
madagra, line 1: Add tests for consistency of wavefunction/expectation value with different `apply_xxx` methods.

madagra (commented):
@dominikandreasseitz can you add the results of your benchmarks as well in the MR description?

tests/test_modules.py
dominikandreasseitz: rm

tests/test_speed.py
dominikandreasseitz: right now, the method is about the same/ sometimes slower than apply_batch_gate which might be due to a lack of optimization of vmap for complex tensors, however we should keep the method for future purposes

madagra:
@dominikandreasseitz before merging, I would just add a simple switch to choose between the `vmap` and the older version of the apply gate function. This can be then used as a configuration parameter in `qadence`.

pyqtorch/modules/parametric.py
```diff
@@ -4,10 +4,9 @@
 from numpy.typing import ArrayLike
 
 from pyqtorch.core.batched_operation import (
-    _apply_batch_gate,
     create_controlled_batch_from_operation,
 )
-from pyqtorch.core.utils import OPERATIONS_DICT
+from pyqtorch.core.utils import OPERATIONS_DICT, _apply_batch_gate
```
madagra, line 9: Where is this used?

nmheim (approved):
Weirdly, I have no complaints! :P

jpmoutinho:
So is it better in the end? And is it always better, or certain scenarios?

nmheim:
> So is it better in the end? And is it always better, or certain scenarios?

I think its better in terms of memory consumption, but true, would be nice to show that somewhere:)

jpmoutinho (approved):

CONTRIBUTING.md
```diff
@@ -14,33 +14,33 @@ Your course of action will depend on your objective, but generally, you should s
...
+Next, navigate to your new pyqtorch fork directory and mark the main pyqtorch repository as the `upstream`:
 
 ```shell
-git remote add upstream https://github.com/pasqal-io/PyQ.git
+git remote add upstream https://github.com/pasqal-io/pyqtorch.git
 ```
 
 ## Setting up your development environment
 
 We recommended to use `hatch` for managing the environments:
```
Roland-djee, line 35:
```suggestion
We recommended to use `hatch` for managing environments:
```

pyqtorch/matrices.py
```diff
@@ -17,9 +17,55 @@
...
+XMAT = torch.tensor([[0, 1], [1, 0]], dtype=DEFAULT_MATRIX_DTYPE)
+YMAT = torch.tensor([[0, -1j], [1j, 0]], dtype=DEFAULT_MATRIX_DTYPE)
+ZMAT = torch.tensor([[1, 0], [0, -1]], dtype=DEFAULT_MATRIX_DTYPE)
+SMAT = torch.tensor([[1, 0], [0, 1j]], dtype=DEFAULT_MATRIX_DTYPE)
+SDAGGERMAT = torch.tensor([[1, 0], [0, -1j]], dtype=DEFAULT_MATRIX_DTYPE)
+TMAT = torch.tensor(
+    [[1, 0], [0, torch.exp(torch.tensor(1.0j * torch.pi / 4))]], dtype=DEFAULT_MATRIX_DTYPE
+)
+NMAT = torch.tensor([[0, 0], [0, 1]], dtype=DEFAULT_MATRIX_DTYPE)
+NVEC = torch.tensor([0, 1], dtype=DEFAULT_MATRIX_DTYPE)
```
Roland-djee, line 34:
```suggestion
NDIAG = torch.tensor([0, 1], dtype=DEFAULT_MATRIX_DTYPE)
```

pyqtorch/modules/ops.py
```diff
@@ -0,0 +1,53 @@
...
+        state = torch.tensordot(mat, state, dims=axes)
+        inv_perm = torch.argsort(
+            torch.tensor(
+                state_dims + [j for j in range(n_qubits) if j not in state_dims], dtype=torch.int
+            )
+        )
+        state = torch.permute(state, tuple(inv_perm))
+        return state
+
+    return torch.vmap(
```
Roland-djee, line 49: What's the point of doing torch.vmap ? Performance reasons ?

Roland-djee (approved):
LGTM. Few comments for the sake of discussion.

pyqtorch/modules/ops.py
dominikandreasseitz: apply_gate doesnt work on batches/ only works on 2D tensors, _vmap_gate performs apply_gate on a 3D tensor

Roland-djee (approved):

PR #98, opened 2023-08-17
Adjust apply_gate Endianness & Basis State Ordering

dominikandreasseitz:


pyqtorch/core/utils.py
```diff
@@ -119,7 +119,7 @@ def _apply_einsum_gate(
     Same shape as `ìnput_state`
     """
     mat = mat.reshape([2] * len(qubits) * 2)
-    qubits = N_qubits - 1 - np.array(qubits)
+    qubits = qubits
```
madagra, line 122: Not needed.

pyqtorch/core/utils.py
```diff
@@ -86,23 +85,20 @@ def _apply_gate(
     - state (torch.Tensor): the quantum state after application of the gate.
     Same shape as `ìnput_state`
     """
-    mat = mat.reshape([2] * len(qubits) * 2)
+    mat = mat.view([2] * len(qubits) * 2)
```
nmheim, line 88: Nice this prevents a copy?

nmheim (approved):

pyqtorch/core/utils.py
awennersteen, line 88: why view over reshape?

awennersteen: more like, when reshape would have copied view will fail IIRC

awennersteen:
Doesnt' this require any changes in the non-module api?

I'm not sure everything there is tested at all, e.g. the graph based things. Also should we version bump here whilst we're at it? Or how is this handled in PyQ now?

pyqtorch/core/utils.py
dominikandreasseitz: view just ensures that the tensor is not copied. but i think that .reshape does view under the hood if it can in very recent versions of torch

dominikandreasseitz: yes, mentioned it already in the other thread. probably it didnt copy before but like this we ensure it

awennersteen: I guess for gates we're quite confident that they will always be viewable and contigous, but I'm not sure that it's necessarily the case for Hamiltonians. Are they also using apply_gate? If so, do we think that they'll always be viewable?

This SO post agrees with what I remember:
https://stackoverflow.com/questions/49643225/whats-the-difference-between-reshape-and-view-in-pytorch

dominikandreasseitz:
@awennersteen
_Doesnt' this require any changes in the non-module api?_ : good question, i contacted @Slimane33 about this and lets see what he says
_bump version_ : yes i will/have

pyqtorch/core/utils.py
dominikandreasseitz: they use _apply_batch_gate which uses `einsum` and that reshapes under the hood, not sure via allocating copies though, if it can avoid it, then yes but i guess with varying qubit support it has to rearrange the tensor which makes it not contiguous anymore, thats btw a big argument for not using einsum

Slimane33:
Hello, I think the main reason to keep the ordering of qubits as it is is that sometimes you are interested by flattening the state and look at the individual entries. Maybe then we can add a util function to do this permutation on the state directly ?

pyqtorch/modules/utils.py
```diff
@@ -154,3 +155,27 @@ def _rand(n_qubits: int) -> torch.Tensor:
 
     _state = torch.concat(tuple(_rand(n_qubits) for _ in range(batch_size)), dim=1)
     return _state.reshape([2] * n_qubits + [batch_size])
+
+
+def flatten_wf(wf: torch.Tensor) -> torch.Tensor:
+    return torch.flatten(wf, start_dim=0, end_dim=-2).t()
+
+
+def invert_endianness(wf: torch.Tensor) -> torch.Tensor:
```
dominikandreasseitz, line 164: @Slimane33 i added the function to convert the wf here

awennersteen (approved):
lgtm

pyqtorch/core/utils.py
```diff
@@ -86,23 +85,20 @@ def _apply_gate(
...
+    qubits = list(qubits)
+    axes = (mat_dims, qubits)
     state = torch.tensordot(mat, state, dims=axes)
     inv_perm = torch.argsort(
-        torch.tensor(
-            state_dims + [j for j in range(N_qubits + 1) if j not in state_dims], dtype=torch.int
-        )
+        torch.tensor(qubits + [j for j in range(N_qubits + 1) if j not in qubits], dtype=torch.int)
     )
     state = torch.permute(state, tuple(inv_perm))
```
jpmoutinho, line 96: What is this permutation doing?

dominikandreasseitz: when you do `tensordot` on a subset of dims on two tensors, the contracted dimensions are added as the first dimensions of the tensor, the permute reverts this back to the original order

jpmoutinho (approved):

PR #106, opened 2023-09-15
Add kjd_gate

dominikandreasseitz:


dominikandreasseitz:
this is mostly for benchmarking purposes and to see if tensordot increases the memory consumption with repeated application (in contrast to normal matmul). note that this function requires the allocation of the indices which are huge long tensors

dominikandreasseitz:
@Roland-djee, the method in this MR can be used to contract over target qubits, maybe its useful for #113

dominikandreasseitz:
merged with https://github.com/pasqal-io/pyqtorch/pull/125/files

PR #107, opened 2023-10-02
[Feat] Adjoint differentiation

dominikandreasseitz:
Implementation of https://arxiv.org/abs/2009.02823

- [x] Add dagger, jacobian, unitary matrices methods
- [x] Add adjoint forward/backward pass
- [x] Add option to pass param_dict: dict[str,torch.Tensor] to QC and gates
- [x] Add tests
- [ ] Add higher-order diff https://pytorch.org/tutorials/intermediate/custom_function_double_backward_tutorial.html
- [x] Add 2qubit jacobian
- [x] Associate gates with param names


pyqtorch/modules/adjoint.py
```diff
@@ -0,0 +1,48 @@
...
+    @staticmethod
+    def backward(ctx: Any, grad_out: Tensor) -> tuple:
+        param_values = ctx.saved_tensors
+        values = param_dict(ctx.param_names, param_values)
+        grads: list = []
+        for op in ctx.circuit.reverse().operations:
+            ctx.out_state = op.apply_dagger(ctx.out_state, values)
+            if isinstance(op, ParametricGate):
+                mu = op.apply_jacobian(ctx.out_state, values)
+                grads = [grad_out * 2 * pyq.overlap(ctx.projected_state, mu)] + grads
```
madagra, line 44: From where does the factor `2` come from exactly?

pyqtorch/modules/adjoint.py
```diff
@@ -0,0 +1,48 @@
...
+            else:
+                grads = [None] + grads
+            ctx.projected_state = op.apply_dagger(ctx.projected_state, values)
```
madagra, line 47: I guess this should enable the application of higher-order derivatives without problems, right? Did you test it?

madagra (commented):
Nice implementation, @dominikandreasseitz. I have a few comments/clarifications. Am I correct that it is working currently only for single qubit rotations?

dominikandreasseitz:
closing since added in the remake

PR #109, opened 2023-10-05
Use custom docker image for CI

dominikandreasseitz:


.github/workflows/run-tests-and-mypy.yml
```diff
@@ -13,10 +13,10 @@ jobs:
     if: |
       github.event_name == 'push' || github.event_name == 'pull_request'
     name: Unit testing
-    runs-on: ubuntu-latest
+    runs-on: docker.io/pytorch/torchserve
```
madagra, line 16: @dominikandreasseitz this is not exactly the way you should use the image. It should be at the job level (now I remembered). Basically:

```
   jobs:
        tests:
            runs-on: ubuntu-18.04
            container: 
                image: docker.io/pytorch/torchserve
```

See here for more details: https://stackoverflow.com/questions/64033686/how-can-i-use-private-docker-image-in-github-actions.

PR #118, opened 2023-10-16
Prototype 2x2 matmul function to be compiled.

### LINKED ISSUE #113: Make model compilation work
## Motivations

PyTorch 2.0 provides a [compilation routine](https://pytorch.org/tutorials/intermediate/torch_compile_tutorial.html) that JIT-compile PyTorch code into optimized kernels, similarly to what Jax does. This is a very simple way to significantly speed up PyTorch code. However, currently the `torch.compile` routine does not have support for complex numbers and it cannot be applied directly to PyQTorch.

## Proposal

In order to be able to use `torch.compile` with models created using PyQTorch, we should remove complex numbers from the library. The fastest way to do so is to split real and imaginary parts of the wavefunction into two real-valued tensors which are treated separately in the computation and put together only when requested.

To keep in mind:
- [x] make real/imaginary parts separation
- [x] implement simple interface to allow for compilation, user should not be involved ideally
  - [x] Implemented, compiled and benchmarked for 2x2 matmuls and 2x2 matvecmuls
  - [ ] Implemented, compiled and benchmarked for tensordots
  - [ ] Implemented, compiled and benchmarked for einsum
- [x] benchmark implementation with current complex-value approach
- [ ] remove the whole complex-valued code if performance is acceptable, otherwise maintain it alongside the real-valued one

Roland-djee:
Fixes #113.

### DIFF
--- a/pyqtorch/modules/matrix.py
+++ b/pyqtorch/modules/matrix.py
@@ -1,36 +1,68 @@
+from __future__ import annotations
+
 import torch
 from torch import Tensor
+
 
 class CustomMatrix(Tensor):
     """Prototype for a custom 2x2 matmul."""
 
     def __matmul__(self, other: object):
-        R11 = torch.matmul(self.real[0,:], other.real[:,0]) - torch.matmul(self.imag[0,:], other.imag[:,0])
-        R12 = torch.matmul(self.real[0,:], other.real[:,1]) - torch.matmul(self.imag[0,:], other.imag[:,1])
-        R21 = torch.matmul(self.real[1,:], other.real[:,0]) - torch.matmul(self.imag[1,:], other.imag[:,0]) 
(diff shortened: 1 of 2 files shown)


Roland-djee:
> Hey @dominikandreasseitz. Here is a prototype for a 2x2 matrix matmul by separating real/imag parts. Question: does it work straight out of the box for you ? I've got clang warnings that it doesn't find `torch/include`.

OK I could make it work in a conda env.

dominikandreasseitz:
great, lets try with tensordot and 2qubit gates

nmheim:
Maybe it makes sense to look at what’s done here https://github.com/albanD/subclass_zoo/blob/main/complex_tensor.py

Roland-djee:
@dominikandreasseitz closing as stale for too long.

PR #125, opened 2023-11-06
[Proto] Mutate state

### LINKED ISSUE #165: [Performance] Mutate state instead of using einsum for non-parametric operations
For non-parametric operations, we use tensordot/einsum to apply the operator. However, it could be more computationally efficient to simply mutate the existing state tensor

Prototype in: https://github.com/pasqal-io/pyqtorch/pull/125

Edit by @jpmoutinho :

The ideas behind this are described in https://arxiv.org/pdf/2303.01493

An example (which was what we initially did in the prototype above) is the `X` gate: it should target all the state amplitudes where the target qubit is 0 and exchange their place with the ones where it is 1. Other rules for each gate have to be implemented.

The implementation should essentially amount to overriding the `forward` method of each gate in `pyqtorch.primitives.primitive_gates` and `pyqtorch.primitives.parametric_gates` (well, the ones for which a rule can be added).

Testing it should be easy: just comparing the result to the normal usage of the `apply_operator`. The function `calc_mat_vec_wavefunction` in `tests.helpers` would be useful. It is used extensively in the `test_tensor.py`.

Overall, this could be interesting to explore, but it is not clear if we can make it efficient in torch (Spinoza does it in Rust). It should only be implemented if there is some clear performance gain over the normal `apply_operator`.

dominikandreasseitz:
Closes #165

### DIFF
--- a/pyqtorch/apply.py
+++ b/pyqtorch/apply.py
@@ -3,11 +3,12 @@ from __future__ import annotations
 from string import ascii_letters as ABC
 from typing import Tuple
 
+import torch
 from numpy import array
 from numpy.typing import NDArray
 from torch import einsum
 
-from pyqtorch.utils import Operator, State
+from pyqtorch.utils import Operator, State, basis_state_indices
 
 ABC_ARRAY: NDArray = array(list(ABC))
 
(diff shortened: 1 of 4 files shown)


Roland-djee:
@dominikandreasseitz Can it be marked as stale or closed ?

dominikandreasseitz:
@Roland-djee, making this idea work will greatly improve the performance of pyq. i am planning to do it at some point if i have time

Roland-djee:
> @Roland-djee, making this idea work will greatly improve the performance of pyq. i am planning to do it at some point if i have time

OK great. We will need to compile all backends related issues at some point. So that we trace them a bit better.

dominikandreasseitz:
@jpmoutinho lets not forget about this branch, we could potentially gain a lot of performance here. however out of scope for me atm. assigning it to you

PR #126, opened 2023-11-07
[Fix] Matrix precision and use no_grad in adjoint

dominikandreasseitz:
Added missing dtype to matrix inits which causes issues in qadence. added no_grad decorator to first order adjoint, bump version


jpmoutinho (approved):

Roland-djee (approved):
Thanks !

PR #127, opened 2023-11-09
Allow for python 3.12

awennersteen:


awennersteen:
Not possible to run torch on 3.12

PR #135, opened 2024-01-31
[Docs / Fix] Add documentation + Correct GPU Device inference

dominikandreasseitz:
- Remake of pyq gates into pure torch.nn.Modules enables automatic moving of buffers to device. 
- Adding a custom 'to' method to QuantumCircuit since without it, the gates are not moved to the device


pyqtorch/circuit.py
```diff
@@ -74,15 +75,29 @@ def expectation(
...
+                device_lst += [b.device for b in op.buffers()]
+        maybe_device = list(set(device_lst))
+        if len(maybe_device) > 1:
+            logger.error(f"Found more than one device in module {self}.")
+        elif len(maybe_device) == 0:
+            logger.warning(f"Unable to determine device of module {self}.")
+        else:
+            _device = device_lst[0]
+            logger.info(f"Found device {_device}.")
+            return _device
```
awennersteen, line 93: How often would this be called?

I think maybe `logger.debug` would be more appropriate.


also, you do an error log but instead of raising you return None. I think that should either not be logged error, or raise an exception.

I'd suggest that instead you rewrite as
```python
if len(maybe_device) == 1:
      ...
 else:
        logger.warning(f"Unable to determine device of module {self}. Found {maybe_device}, expected exactly one device")
```

pyqtorch/circuit.py
```diff
@@ -74,15 +75,29 @@ def expectation(
...
-
-    def init_state(self, batch_size: int = 1) -> torch.Tensor:
+    def _device(self) -> device:
+        device_lst: list[device] = []
+        for op in self.operations:
+            if isinstance(op, QuantumCircuit):
+                device_lst.append(op._device)
+            elif isinstance(op, Module):
+                device_lst += [b.device for b in op.buffers()]
+        maybe_device = list(set(device_lst))
```
awennersteen, line 85: why not a set from the beginning?
I think the rest of the code will also work with sets too?

Although I'm a bit concerned if `b.device` could be a str? or would it by then be guaranteed to be a torch.device? set doesnt necessarily behave as expected with strings

pyqtorch/circuit.py
dominikandreasseitz: same answer as below. once for determining the device on which the module buffers sit and then initializing  an initial state on the same device

pyqtorch/circuit.py
awennersteen: yes, that I understand.
But why use first a `list`, then a `set`, and then a `list` again?
I think you'll be fine with a single `set` the whole time.

pyqtorch/circuit.py
awennersteen: yes, that I understand. 
I'll rewrite for clarity

awennersteen: I was on my phone and ended up with multiple questions in a single comment, sorry about that.

pyqtorch/circuit.py
dominikandreasseitz: good point

dominikandreasseitz: its a set now

dominikandreasseitz:
```
In [1]: myset = set()

In [2]: myset.add('bla')

In [3]: myset.add('bla')

In [4]: myset
Out[4]: {'bla'}
```

pyqtorch/circuit.py
dominikandreasseitz: rewritten to match your suggestions

docs/index.md
```diff
@@ -169,15 +171,10 @@ import matplotlib.pyplot as plt
...
 
-
-def target_function(x: torch.Tensor, degree: int = 3) -> torch.Tensor:
-    result = 0
-    for i in range(degree):
-        result += torch.cos(i*x) + torch.sin(i*x)
-    return .05 * result
-
+# A target function and some training data
+fn = lambda x, degree: .05 * reduce(add, (torch.cos(i*x) + torch.sin(i*x) for i in range(degree)), 0)
```
awennersteen, line 177: Very nice line of code, but I fear it might be confusing for users?

pyqtorch/circuit.py
```diff
@@ -43,46 +43,71 @@ def __eq__(self, other: Any) -> bool:
...
+        raise ValueError("Please provide an observable to compute expectation.")
+    if state is None:
+        state = circuit.init_state(batch_size=1)
+    if diff_mode == DiffMode.AD:
+        state = circuit.run(state, values)
+        return overlap(state, observable.forward(state, values))
+    else:
+        from pyqtorch.adjoint import AdjointExpectation
+
+        return AdjointExpectation.apply(circuit, observable, state, values.keys(), *values.values())
```
awennersteen, line 113: if the user erroneously passes `DiffMode.GPSR` this will not result in the expected outcome. 
I'd say in that case you also raise `ValueError`.

tests/test_circuit.py
```diff
@@ -78,3 +77,10 @@ def _fwd(phi: torch.Tensor, theta: torch.Tensor, epsilon: torch.Tensor) -> torch
         return circ(state, {"phi": phi, "theta": theta, "epsilon": epsilon})
 
     assert torch.autograd.gradcheck(_fwd, (phi, theta, epsilon))
+
+
+def test_device_inference() -> None:
+    ops = [pyq.RX(0), pyq.RX(0)]
+    circ = pyq.QuantumCircuit(2, ops)
+    nested_circ = pyq.QuantumCircuit(2, [circ, circ])
+    assert nested_circ._device is not None
```
awennersteen, line 86: I guess this is then a case that failed in the past?

pyqtorch/circuit.py
awennersteen: neat

pyqtorch/circuit.py
awennersteen: thanks!

tests/test_circuit.py
dominikandreasseitz: no but i just wanted to have a test in the suite which consistently checks that this property correctly works

awennersteen (approved):
lgtm.

Up to you what you think is better with the function

PR #140, opened 2024-02-22
[Proto] Data, Model and Pipeline Parallelism example

### LINKED ISSUE #164: [Feature] Data, Model and Pipeline Parallelism example
@dominikandreasseitz Could you please provide a description ?

dominikandreasseitz:
Closes #164

### DIFF
--- a/examples/parallel.py
+++ b/examples/parallel.py
@@ -1,0 +1,55 @@
+from __future__ import annotations
+
+import torch
+
+import pyqtorch as pyq
+from pyqtorch.parametric import Parametric
+
+N_DEVICES = 2
+
+assert torch.cuda.device_count() == N_DEVICES
+
+
+class ParallelCircuit(torch.nn.Module):
+    def __init__(self, c0: pyq.QuantumCircuit, c1: pyq.QuantumCircuit, params_c0, params_c1):
+        super().__init__()
+        self.c0 = c0.to("cuda:0")
+        self.c1 = c1.to("cuda:1")
+        self.params_c0 = torch.nn.ParameterDict({k: v.to("cuda:0") for k, v in params_c0.items()})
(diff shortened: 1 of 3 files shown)


dominikandreasseitz:
@awennersteen it looks like i wont be able to finish this MR, but it seems like something you would enjoy doing so please either try to finish the task yourself or pass it to someone

Roland-djee:
> @awennersteen it looks like i wont be able to finish this MR, but it seems like something you would enjoy doing so please either try to finish the task yourself or pass it to someone

I can also give it a try jointly with @awennersteen

Roland-djee:
Hey @awennersteen do you think we can merge and close it ?

awennersteen:
Hi @Roland-djee 

I forgot to respond here when it happened.

I remember the issue here more or less. It was an clever idea Dominik had to use multiple GPUs.
Frankly this was only clear to Dominik and I'm not entirely sure how to do this in a way where we really gain something. 
The classical NN equivalent is a lot simpler to parallelize in this way, and it was never 100% clear if possible for quantum


That being difficult is part of the reason why it's stuck here.
Let me try to have a look at this now in August and I'll check if there is anything we'd really like to keep.

Roland-djee:
> Hi @Roland-djee
> 
> I forgot to respond here when it happened.
> 
> I remember the issue here more or less. It was an clever idea Dominik had to use multiple GPUs. Frankly this was only clear to Dominik and I'm not entirely sure how to do this in a way where we really gain something. The classical NN equivalent is a lot simpler to parallelize in this way, and it was never 100% clear if possible for quantum
> 
> That being difficult is part of the reason why it's stuck here. Let me try to have a look at this now in August and I'll check if there is anything we'd really like to keep.

Cool, thanks a lot !

chMoussa:
Closing as obsolete

Issue #141, opened 2024-02-29
Enable setting dtype via torch.set_default_dtype


dominikandreasseitz:

dominikandreasseitz:
closing after #142

PR #142, opened 2024-02-29
[Feature] Allow setting precision

dominikandreasseitz:
https://github.com/pasqal-io/pyqtorch/issues/141


pyqtorch/utils.py
```diff
@@ -5,11 +5,14 @@
...
 import torch
 
-from pyqtorch.matrices import DEFAULT_MATRIX_DTYPE
+from pyqtorch.matrices import DEFAULT_MATRIX_DTYPE, DEFAULT_REAL_DTYPE
 
 State = torch.Tensor
 Operator = torch.Tensor
 
+ATOL = 1e-07 if DEFAULT_REAL_DTYPE == torch.float32 else 1e-014
+GRADCHECK_ATOL = 1e-01 if DEFAULT_REAL_DTYPE == torch.float32 else 1e-014
```
awennersteen, line 14: that's a huge difference?

pyqtorch/matrices.py
```diff
@@ -3,7 +3,8 @@
 import torch
 from torch import Tensor
 
-DEFAULT_MATRIX_DTYPE = torch.cdouble
+DEFAULT_REAL_DTYPE = torch.get_default_dtype()
+DEFAULT_MATRIX_DTYPE = torch.cdouble if DEFAULT_REAL_DTYPE == torch.float64 else torch.cfloat
```
awennersteen, line 7: This always makes me afarid of order of execution issues.

Is it not better to query for this when needed? Rather than here which will probably run on `import pyqtorch`?

pyqtorch/utils.py
dominikandreasseitz: yeah but grad_check fails otherwise, we could not use gradcheck and do it manually like in other tests

awennersteen: I guess let's keep it like this. 
As you pointed out in the DM it seems to be standard. I just didn't except e-14 -> e-1...

dominikandreasseitz: i think this is definitely a failure on the torch side

pyqtorch/matrices.py
dominikandreasseitz: jep, good point

dominikandreasseitz: not sure how to realize this exactly. i'd say that if the user wants to use higher precision, they should always put the `torch.set_default_dtype` call at the very start of their script.

tests/test_digital.py
```diff
@@ -105,26 +105,26 @@ def test_CNOT_state00_controlqubit_0() -> None:
 
 def test_CNOT_state10_controlqubit_0() -> None:
     result: torch.Tensor = pyq.CNOT(0, 1)(product_state("10"), None)
-    assert torch.equal(product_state("11"), result)
+    assert torch.allclose(product_state("11"), result)
```
jpmoutinho, line 108: Missing `atol`? And same comment about the `rtol` I made above.

tests/test_digital.py
```diff
@@ -221,7 +221,7 @@ def test_parametric_phase_hamevo(
 def test_parametrized_phase_gate(state_fn: Callable, batch_size: int, n_qubits: int) -> None:
     target: int = torch.randint(low=0, high=n_qubits, size=(1,)).item()
     state = state_fn(n_qubits, batch_size=batch_size)
-    phi = torch.tensor([torch.pi / 2], dtype=torch.cdouble)
+    phi = torch.tensor([torch.pi / 2], dtype=DEFAULT_MATRIX_DTYPE)
     phase = pyq.PHASE(target, "phi")
     constant_phase = pyq.S(target)
     assert torch.allclose(phase(state, {"phi": phi}), constant_phase(state, None))
```
jpmoutinho, line 227: Same about `atol, rtol`.

jpmoutinho (commented):
Thanks @dominikandreasseitz!

So in principle this will simply grab the default `dtype` from Qadence? But how does it behave if a user sets the default `dtype` inside a script after importing Qadence?

Roland-djee (approved):
Nice. LGTM.

pyqtorch/utils.py
```diff
@@ -5,11 +5,15 @@
...
 import torch
 
-from pyqtorch.matrices import DEFAULT_MATRIX_DTYPE
+from pyqtorch.matrices import DEFAULT_MATRIX_DTYPE, DEFAULT_REAL_DTYPE
 
 State = torch.Tensor
 Operator = torch.Tensor
 
+ATOL = 1e-07 if DEFAULT_REAL_DTYPE == torch.float32 else 1e-014
+RTOL = 1e-05 if DEFAULT_REAL_DTYPE == torch.float32 else 0.0
```
jpmoutinho, line 14: Why do you need the `1e-05` rtol for float32 though? If tests are not passing otherwise, maybe it would be sufficient to make the default ATOL `1e-06` for that case. 

[According to this](https://en.wikipedia.org/wiki/Machine_epsilon) the machine epsilon for float32 can be anywhere from `5e-08` to `1e-07`, so probably we are pushing it too close with the current value.

jpmoutinho (approved):

pyqtorch/utils.py
dominikandreasseitz: its manually tested using 1e-06 now

examples/dqc.py
```diff
@@ -0,0 +1,164 @@
...
+X_POS = 0
+Y_POS = 1
+N_POINTS = 150
+N_EPOCHS = 1000
+
+
+def hea(n_qubits: int, n_layers: int, param_name: str) -> list:
+    ops = []
+    for layer in range(n_layers):
+        ops += [pyq.RX(i, f"{param_name}_0_{layer}_{i}") for i in range(n_qubits)]
```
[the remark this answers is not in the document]
dominikandreasseitz: done

examples/dqc.py
```diff
@@ -0,0 +1,164 @@
...
+        ops += [pyq.RX(i, f"{param_name}_2_{layer}_{i}") for i in range(n_qubits)]
+        ops += [pyq.CNOT(i % n_qubits, (i + 1) % n_qubits) for i in range(n_qubits)]
+    return ops
+
+
+class TotalMagnetization(pyq.QuantumCircuit):
+    def __init__(self, n_qubits: int):
+        super().__init__(n_qubits, [pyq.Z(i) for i in range(n_qubits)])
+
+    def forward(self, state, values) -> torch.Tensor:
```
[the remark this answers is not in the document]
dominikandreasseitz: done

pyqtorch/utils.py
dominikandreasseitz: done

docs/index.md
```diff
@@ -237,3 +237,184 @@ def fig_to_html(fig: Figure) -> str:  # markdown-exec: hide
...
+ansatz = hea(N_QUBITS, DEPTH, "theta")
+param_dict = torch.nn.ParameterDict(
+    {
+        op.param_name: torch.rand(1, requires_grad=True)
+        for op in ansatz
+        if isinstance(op, Parametric)
+    }
+)
+circ = QuantumCircuit(N_QUBITS, feature_map + ansatz).to(device=DEVICE, dtype=COMPLEX_DTYPE)
+observable = TotalMagnetization(N_QUBITS).to(device=DEVICE, dtype= COMPLEX_DTYPE)
```
Roland-djee, line 363:
```suggestion
observable = TotalMagnetization(N_QUBITS).to(device=DEVICE, dtype=COMPLEX_DTYPE)
```

pyqtorch/adjoint.py
```diff
@@ -33,23 +32,26 @@ def forward(
...
+                    else:
+                        grad = zeros(1)
+
+                    grads_dict[op.param_name] = grad
+
+                ctx.projected_state = apply_operator(
+                    ctx.projected_state, op.dagger(values), op.qubit_support
+                )
+            except Exception as e:
+                breakpoint()
```
Roland-djee, line 56: Is it some debugging left over ?

pyqtorch/circuit.py
```diff
@@ -64,15 +65,24 @@ def forward(self, state: State, values: dict[str, Tensor] = {}) -> State:
...
 
-    def to(self, device: torch_device) -> QuantumCircuit:
-        self.operations = ModuleList([op.to(device) for op in self.operations])
-        self._device = device
+    def to(self, *args: Any, **kwargs: Any) -> QuantumCircuit:
+        self.operations = ModuleList([op.to(*args, **kwargs) for op in self.operations])
+        try:
+            self._device = self.operations[0].device
+            self._dtype = self.operations[0].dtype
+        except IndexError:
```
Roland-djee, line 83: Wouldn't it be preferrable to check for `self.operations` emptiness ?

pyqtorch/matrices.py
```diff
@@ -3,6 +3,7 @@
 import torch
 from torch import Tensor
 
+DEFAULT_REAL_DTYPE = torch.float64
```
Roland-djee, line 6: It doesn't seem to be used ?

Roland-djee (approved):
Thanks @dominikandreasseitz. Not much to say. LGTM.

pyqtorch/matrices.py
dominikandreasseitz: redone now so you pass the dtype like a device

pyqtorch/utils.py
```diff
@@ -41,19 +44,21 @@ class DiffMode(StrEnum):
...
 
 
-def is_normalized(state: torch.Tensor, atol: float = 1e-14) -> bool:
+def is_normalized(state: torch.Tensor, atol: float = ATOL) -> bool:
     n_qubits = len(state.size()) - 1
     batch_size = state.size()[-1]
     state = state.reshape((2**n_qubits, batch_size))
     sum_probs = (state.abs() ** 2).sum(dim=0)
-    ones = torch.ones(batch_size, dtype=torch.double)
-    return torch.allclose(sum_probs, ones, rtol=0.0, atol=atol)  # type: ignore[no-any-return]
```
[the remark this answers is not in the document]
dominikandreasseitz: using all atol=1e-06 and rtol=0.0  now like you suggested

pyqtorch/adjoint.py
dominikandreasseitz: yes, good catch

pyqtorch/matrices.py
dominikandreasseitz: it is used in various test files etc

pyqtorch/circuit.py
dominikandreasseitz: true

Issue #145, opened 2024-03-08
[Feature] Add Kraus Operator


EthanObadia:
We know that the state of an open quantum system is represented by the density matrix $\rho$. But how do we describe the dynamics of this system?

The transformations of the density operator of an open quantum system interacting with its environment are represented by the super-operator $S: \rho \rightarrow S(\rho)$, often referred to as a quantum channel. 
Quantum channels, due to the conservation of the probability distribution, must be CPTP (Completely Positive and Trace Preserving). Any CPTP super-operator can be written in the following form:

$$
S(\rho) = \sum_i K_i \rho K^{\dagger}_i 
$$

Where $K_i$ are the Kraus operators, and satisfy the property $\sum_i K_i K^{\dagger}_i = \mathbb{I}$

It is therefore possible to simulate the dynamics of an open quantum system if the Kraus decomposition of the applied quantum channel is known. 
With this in mind, we will implement Kraus operators in Pyqtorch to study the effect of certain noisy quantum channels whose Kraus decomposition has been analytically demonstrated, such as the `depolarizing channel`.

Roland-djee:
Is this closeable @EthanObadia ?

EthanObadia:
> Is this closeable @EthanObadia ?

It will be automatically closed when #148 will be merged (soon).

PR #148, opened 2024-03-19
[Feature] Adding noise gate 

EthanObadia:
Add Noise.py and Noise class for implementing the Kraus Operator of well known noisy gates (#145).

### DIFF
--- a/pyqtorch/noise.py
+++ b/pyqtorch/noise.py
@@ -1,0 +1,5 @@
+from __future__ import annotations
+
+import torch
+
+from pyqtorch.utils import Operator, State
--- a/pyqtorch/utils.py
+++ b/pyqtorch/utils.py
@@ -127,7 +127,7 @@ def density_mat(vector_state:State) -> Operator :
     density_mat = torch.zeros(2**n_qubits,2**n_qubits,batch_size,dtype=DEFAULT_MATRIX_DTYPE)
     for i in range (batch_size):
         v_state_ket = vector_state.reshape((2**n_qubits, batch_size))[:,i].reshape(2**n_qubits,1)
-        v_state_bra = v_state_ket.transpose(1,0)
+        v_state_bra = v_state_ket.transpose(1,0).conj()
         density = torch.matmul(v_state_ket,v_state_bra)
         density_mat[:,:,i] = density
     return density_mat
--- a/pyqtorch/utils.py
+++ b/pyqtorch/utils.py
@@ -116,3 +116,18 @@ def random_state(
 
 def param_dict(keys: Sequence[str], values: Sequence[torch.Tensor]) -> dict[str, torch.Tensor]:
     return {key: val for key, val in zip(keys, values)}
+
+
+def density_mat(vector_state:State) -> Operator :
+    """
+    Transforms a pure state :math:`|\\psi\\rangle` into its corresponding density matrix :math:`\\rho = |\\psi\\rangle \\langle\\psi|`.
+    """
+    n_qubits = len(vector_state.size()) - 1
+    batch_size = vector_state.size(-1)
+    density_mat = torch.zeros(2**n_qubits,2**n_qubits,batch_size,dtype=DEFAULT_MATRIX_DTYPE)
+    for i in range (batch_size):
+        v_state_ket = vector_state.reshape((2**n_qubits, batch_size))[:,i].reshape(2**n_qubits,1)
+        v_state_bra = v_state_ket.transpose(1,0)
+        density = torch.matmul(v_state_ket,v_state_bra)
+        density_mat[:,:,i] = density
+    return density_mat


pyqtorch/noise.py
```diff
@@ -0,0 +1,83 @@
...
+
+from typing import Tuple
+
+import torch
+
+from pyqtorch.matrices import DEFAULT_MATRIX_DTYPE, IMAT, XMAT, _dagger
+from pyqtorch.utils import Operator, State
+
+
+class Noise(torch.nn.Module):
```
dominikandreasseitz, line 11: can we make this class inherit from `Primitive`?

pyqtorch/apply.py
```diff
@@ -57,3 +60,35 @@ def apply_operator(
         map(lambda e: "".join(list(e)), [operator_dims, in_state_dims, out_state_dims])
     )
     return einsum(f"{operator_dims},{in_state_dims}->{out_state_dims}", operator, state)
+
+
+def apply_ope_ope(ope1: torch.Tensor, ope2: torch.Tensor) -> torch.Tensor:
```
dominikandreasseitz, line 65: to me, it seems we can simply do a batch matmul here.however you need to move the batch dim to the first dimension so torch doesnt complain.
`   
        batch_first_perm = (2, 0, 1)
        undo_perm = tuple(argsort(tensor(batch_first_perm)))

        def _batch_first(m: Tensor) -> Tensor:
            return permute(m, batch_first_perm)  # This returns shape (batch_size, 2, 2)

        def _batch_last(m: Tensor) -> Tensor:
            return permute(
                m, undo_perm
            )  # We need to undo the permute since PyQ expects (2, 2, batch_size).

        return _batch_last(torch.bmm(_batch_first(ope1), _batch_first(ope2) )
        `

EthanObadia: What do you think of this version ?

```python 
def apply_ope_ope(operator_1: torch.Tensor, operator_2: torch.Tensor) -> torch.Tensor:
    """
    Compute the product of two operators.   
    The operators must have the same dimensions.

    Args:
        operator_1 (torch.Tensor): The first operator.
        operator_2 (torch.Tensor): The second operator.

    Returns:
        torch.Tensor: The product of the two operators.

    Raises:
        ValueError: If the number of qubits or the batch size differs between the two operators.
    """

    # Dimension verifications:
    n_qubits_1 = int(log2(operator_1.size(1)))
    n_qubits_2 = int(log2(operator_2.size(1)))
    batch_size_1 = operator_1.size(-1)
    batch_size_2 = operator_2.size(-1)
    if n_qubits_1 != n_qubits_2:
        raise ValueError("The number of qubit is different between the two operators.")
    if batch_size_1 != batch_size_2:
        raise ValueError("The number of batch is different between the two operators.")
    
    # Permute the batch size on first dimension to allow torch.bmm():
    def batch_first(operator: torch.Tensor) -> torch.Tensor:
        """
        Permute the operator's batch dimension on first dimension.

        Args:
        operator (torch.Tensor): Operator in size [2**n_qubits, 2**qubits,batch_size].

        Returns:
        torch.Tensor: Operator in size [batch_size, 2**n_qubits, 2**qubits].
        """
        batch_first_perm = (2, 0, 1)
        return torch.permute(operator, batch_first_perm) 

    # Undo the permute since PyQ expects tensor.Size([2**n_qubits, 2**qubits,batch_size]):
    def batch_last(operator: torch.Tensor) -> torch.Tensor:
        """
        Permute the operator's batch dimension on last dimension.

        Args:
        operator (torch.Tensor): Operator in size [batch_size,2**n_qubits, 2**qubits].

        Returns:
        torch.Tensor: Operator in size [2**n_qubits, 2**qubits,batch_size].
        """
        undo_perm = (1,2,0)
        return torch.permute(operator, undo_perm) 

    return batch_last(torch.bmm(batch_first(operator_1), batch_first(operator_2)))

File: pyqtorch/noise.py (hunk 1 above)
Comment (EthanObadia) [reply pyqtorch/noise.py:11]: Yes, good idea. I will modify this.

File: pyqtorch/apply.py (hunk 2 above)
Comment (EthanObadia) [reply pyqtorch/apply.py:65]: Yes

File: pyqtorch/noise.py (hunk 1 above)
Comment (EthanObadia) [reply pyqtorch/noise.py:13]: Finally, I don't know because the inputs of the class are `List[Tensor]`. 
In `Primitive`, the `pauli` variable  is a `Tensor`, so everything has to be redefined in `Noise` .
And I don’t know what to write for `pauli` in `super.__init()__`

File: pyqtorch/noise.py [hunk 3]
```diff
@@ -0,0 +1,420 @@
...
+        if not isinstance(kraus, list) or not all(isinstance(tensor, Tensor) for tensor in kraus):
+            raise TypeError("The Kraus operators must be a list containing Tensor objects")
+        if len(kraus) == 0:
+            raise ValueError("The noisy gate must be described by the Kraus operator")
+        # Create a buffer for the Kraus operators.
+        self.kraus = []
+        for index, tensor in enumerate(kraus):
+            # Registering the tensor as a buffer with a unique name
+            self.register_buffer(f"kraus_{index}", tensor)
+            # Storing the tensor itself in the self.kraus list
```
Roland-djee, line 31: Be careful as not to state the obvious in comments. Comments IMHO should describe points that are not expressible in code itself otherwise it makes for an unnecessary line.

pyqtorch/noise.py
```diff
@@ -0,0 +1,420 @@
...
+            self.kraus.append(tensor)
+        # Kraus operator's device
+        if len(set(kraus.device for kraus in self.kraus)) != 1:
```
Roland-djee, line 34: Do you need to check this ? It seems to me that by the very nature of operators in PyQ, everything is in the same device. @dominikandreasseitz ?

pyqtorch/noise.py
```diff
@@ -0,0 +1,420 @@
...
+            raise ValueError("All tensors are not on the same device.")
+        self._device = self.kraus[0].device
+        self.probability: Union[Tuple[float, ...], float] = probability
```
Roland-djee, line 37:
```suggestion
        self.probability: tuple[float, ...] | float = probability
```

Also, renaming suggestion to `probabilities`.

pyqtorch/noise.py
```diff
@@ -0,0 +1,420 @@
...
+        # Verification probability value:
+        if isinstance(self.probability, float):
```
Roland-djee, line 39: If you want to do these kinds of checks, it's probably best to move them the earliest possible.

pyqtorch/noise.py
```diff
@@ -0,0 +1,420 @@
...
+        return self.probability
+
+    @property
+    def name(self) -> str:
+        return self.__class__.__name__
+
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, type(self)):
+            return self.name == other.name and self.probabilty == other.probabilty
+        else:
```
Roland-djee, line 62: Remove.

pyqtorch/noise.py
```diff
@@ -0,0 +1,420 @@
...
+        # Output operator initialization:
+        n_qubits: int = len(state.size()) - 1
+        batch_size: int = state.size(-1)
+        rho_evol: Tensor = torch.zeros(
+            2**n_qubits, 2**n_qubits, batch_size, dtype=DEFAULT_MATRIX_DTYPE
+        )
+
+        # Apply noisy channel on input state
+        rho: Tensor = density_mat(state)
+        kraus_unit: List[Tensor] = self.unitary()
```
Roland-djee, line 112:
```suggestion
        kraus_unitaries: list[Tensor] = self.unitary()
```

pyqtorch/noise.py
```diff
@@ -0,0 +1,420 @@
...
+        kraus_dag: List[Tensor] = self.dagger()
```
Roland-djee, line 113:
```suggestion
        kraus_daggers: list[Tensor] = self.dagger()
```

pyqtorch/noise.py
```diff
@@ -0,0 +1,420 @@
...
+        for i in range(len(self.kraus)):
```
Roland-djee, line 114:
```suggestion
        for kraus_unitary, kraus_dagger in zip(kraus_unitaries, kraus_daggers):
```

pyqtorch/noise.py
```diff
@@ -0,0 +1,420 @@
...
+            rho_i: Tensor = apply_ope_ope(
+                kraus_unit[i], apply_ope_ope(rho, kraus_dag[i], self.target), self.target
+            )
+            rho_evol += rho_i
```
Roland-djee, line 118: Maybe stack the tensors ?

pyqtorch/noise.py
```diff
@@ -0,0 +1,420 @@
...
+    def __init__(
+        self,
+        target: int,
+        probability: float,
+    ):
+        # Verification probability type:
+        if not isinstance(probability, float):
+            raise TypeError("The probability values must be float")
+
+        # Define Kraus operator:
```
Roland-djee, line 157: Maybe something like:

```python
proba_tensor = torch.tensor(probability, dtype=DEFAULT_MATRIX_DTYPE)
```

And then:

```python
K0: Tensor = torch.sqrt(1. - proba_tensor) * IMAT
```

pyqtorch/noise.py
```diff
@@ -0,0 +1,420 @@
...
+        K0: Tensor = torch.sqrt(torch.tensor(1 - probability, dtype=DEFAULT_MATRIX_DTYPE)) * IMAT
+        K1: Tensor = torch.sqrt(torch.tensor(probability / 3, dtype=DEFAULT_MATRIX_DTYPE)) * XMAT
```
Roland-djee, line 228: I would create a tensor object once and reuse it for these tensor probabilities.

pyqtorch/noise.py
```diff
@@ -0,0 +1,420 @@
...
+        TypeError: If the probability values are not provided in a tuple.
+    """
+
+    def __init__(
+        self,
+        target: int,
+        probability: Tuple[float, ...],
+    ):
+        # Verification list probability type:
+        if not isinstance(probability, list):
```
Roland-djee, line 261: Not sure it does the right check.

pyqtorch/primitive.py
```diff
@@ -40,15 +41,15 @@ def extra_repr(self) -> str:
     def param_type(self) -> None:
         return self._param_type
 
-    def unitary(self, values: dict[str, torch.Tensor] | torch.Tensor = {}) -> Operator:
+    def unitary(self, values: dict[str, Tensor] | Tensor = {}) -> Tensor:
```
Roland-djee, line 44:
```suggestion
    def unitary(self, values: dict[str, Tensor] | Tensor = dict()) -> Tensor:
```

Roland-djee (commented):
Thanks @EthanObadia for this preliminary version. Some initial comments form me. Generally: be aware as not to overload the reading with unnecessary comments and watch for the types annotations.

pyqtorch/noise.py
```diff
@@ -0,0 +1,420 @@
...
+from torch import Tensor
+
+from pyqtorch.apply import apply_ope_ope
+from pyqtorch.matrices import DEFAULT_MATRIX_DTYPE, IMAT, XMAT, YMAT, ZMAT, _dagger
+from pyqtorch.utils import density_mat
+
+
+class Noise(torch.nn.Module):
+    def __init__(
+        self, kraus: List[Tensor], target: int, probability: Union[Tuple[float, ...], float]
```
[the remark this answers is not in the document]
EthanObadia: I will change but what is the difference between `Union[Tuple[float, ...], float]` and `tuple[float, ...] | float` and why you prefer to use `tuple[float, ...] | float` ?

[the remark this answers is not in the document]
Roland-djee: Because both the `Union` type and `Tuple` are from `typing` which will be deprecated in the future.

pyqtorch/noise.py
```diff
@@ -0,0 +1,420 @@
...
+    ) -> None:
+        super().__init__()
+        self.target: int = target
+        self.qubit_support: Tuple[int, ...] = (self.target,)
+        # Verification Kraus operators:
+        if not isinstance(kraus, list) or not all(isinstance(tensor, Tensor) for tensor in kraus):
+            raise TypeError("The Kraus operators must be a list containing Tensor objects")
+        if len(kraus) == 0:
```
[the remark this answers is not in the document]
EthanObadia: Erased.

pyqtorch/noise.py
EthanObadia: Changed.

pyqtorch/noise.py
EthanObadia: Noted.

pyqtorch/noise.py
EthanObadia: Done.

pyqtorch/noise.py
EthanObadia: Moved.

pyqtorch/noise.py
EthanObadia: Done.

pyqtorch/noise.py
EthanObadia: Changed.

pyqtorch/primitive.py
EthanObadia: Changed.

pyqtorch/noise.py
EthanObadia: Removed.

pyqtorch/noise.py
```diff
@@ -0,0 +1,420 @@
...
+
+class Noise(torch.nn.Module):
+    def __init__(
+        self, kraus: List[Tensor], target: int, probability: Union[Tuple[float, ...], float]
+    ) -> None:
+        super().__init__()
+        self.target: int = target
+        self.qubit_support: Tuple[int, ...] = (self.target,)
+        # Verification Kraus operators:
+        if not isinstance(kraus, list) or not all(isinstance(tensor, Tensor) for tensor in kraus):
```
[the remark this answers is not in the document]
EthanObadia: Removed.

pyqtorch/noise.py
EthanObadia: Modified.

pyqtorch/noise.py
EthanObadia: Changed.

pyqtorch/noise.py
EthanObadia: Modified.

pyqtorch/noise.py
EthanObadia: I not sure to get it. What's your idea ? What do you to stack and why?

Roland-djee: Well, instead of initialising `rho_evol` as a tensor of zeros and adding on top of that, mayeb use `torch.stack`:

```python
A = torch.stack(A, B) <-- here you stack tensor B to A
```
To be confirmed though.

EthanObadia: Here I am not stacking tensors but performing addition in the idea to compute  $\sum_i K_i \rho K^{\dagger}_i $, right ?
Also `torch.stack` take as input a `tuple[Tensor,...]`: https://pytorch.org/docs/stable/generated/torch.stack.html

EthanObadia: Does this version fit you ?
```python3
    rho_evol = torch.stack(
            [
                operator_product(
                    kraus_unitary, operator_product(rho, kraus_dagger, self.target), self.target
                )
                for kraus_unitary, kraus_dagger in zip(self.unitary(), self.dagger())
            ],
            dim=0,
        )
        rho_evol = torch.sum(rho_evol, dim=0)
        return rho_evol

File: pyqtorch/utils.py [hunk 19]
```diff
@@ -125,6 +125,10 @@ def param_dict(keys: Sequence[str], values: Sequence[Tensor]) -> dict[str, Tenso
     return {key: val for key, val in zip(keys, values)}
 
 
+class Density_Matrix(Tensor):
```
gvelikova, line 128: This is extremely minor but isn't it better to be DensityMatrix instead?

gvelikova (commented):
Very good work @EthanObadia, do you have some agreement with @Roland-djee what tests would make sense to be added here?

EthanObadia:
> Very good work @EthanObadia, do you have some agreement with @Roland-djee what tests would make sense to be added here?
No, I was actually thinking about it. If you have any ideas, I'm all ears.

pyqtorch/utils.py
EthanObadia: I will change.

dominikandreasseitz (commented):
hi @EthanObadia , can you add some documentation?
The way you do that is the following:
1. Create a .md file under `docs/`, just have a look at https://github.com/pasqal-io/pyqtorch/blob/main/docs/index.md to see how its done. 
2. Reference your .md file here https://github.com/pasqal-io/pyqtorch/blob/main/mkdocs.yml#L7 and give it a name, similar to what i did with "pyqtorch in a nutshell" there

To check how your docs look like, execute `hatch run docs:serve` and check your browser. Let me know if something is unclear

dominikandreasseitz (changes requested):
Hi @EthanObadia, can you add some tests + some example in the docs with plots where you use the noise gate

pyqtorch/noise.py
```diff
@@ -0,0 +1,303 @@
...
+            # Storing the tensor itself in the self.kraus list
+            self.kraus.append(tensor)
+
+        # Kraus operator's device
+        if len(set(kraus.device for kraus in self.kraus)) != 1:
+            raise ValueError("All tensors are not on the same device.")
+        self._device = self.kraus[0].device
+
+    @property
+    def proba(self) -> Union[Tuple[float, ...], float]:
```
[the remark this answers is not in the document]
EthanObadia: Changed for `prob`.

pyqtorch/noise.py
```diff
@@ -0,0 +1,303 @@
...
+        Args:
+            kraus_op (Tensor): The single unitary operator tensor.
+
+        Returns:
+            Tensor: A tensor containing batched unitary operators.
+
+        Raises:
+            TypeError: If the input is not a Tensor.
+        """
+        return [kraus_op.unsqueeze(2) for kraus_op in self.kraus]
```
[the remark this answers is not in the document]
EthanObadia: In the current version, the docstring has been removed to follow `Primitive`

pyqtorch/noise.py
```diff
@@ -0,0 +1,303 @@
...
+        """
+        Computes the conjugate transpose (dagger) of a Kraus operator.
+
+        Args:
+            kraus_op (Tensor): The tensor representing a quantum Kraus operator.
+
+        Returns:
+            Tensor: The conjugate transpose (dagger) of the input tensor.
+        """
+        return [_dagger(kraus_op) for kraus_op in self.unitary()]
```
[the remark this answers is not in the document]
EthanObadia: In the current version, the docstring has been removed to follow `Primitive`

pyqtorch/apply.py
```diff
@@ -57,3 +59,63 @@ def apply_operator(
         map(lambda e: "".join(list(e)), [operator_dims, in_state_dims, out_state_dims])
     )
     return einsum(f"{operator_dims},{in_state_dims}->{out_state_dims}", operator, state)
+
+
+def apply_ope_ope(operator_1: Tensor, operator_2: Tensor, target: int) -> Tensor:
```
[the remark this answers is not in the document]
EthanObadia: Changed in the current version.

pyqtorch/noise.py
```diff
@@ -0,0 +1,303 @@
...
+        batch_size: int = state.size(-1)
+        rho_evol: Tensor = torch.zeros(
+            2**n_qubits, 2**n_qubits, batch_size, dtype=DEFAULT_MATRIX_DTYPE
+        )
+
+        # Apply noisy channel on input state
+        rho: Tensor = density_mat(state)
+        kraus_unit = self.unitary()
+        kraus_dag = self.dagger()
+        for index in range(len(self.kraus)):
```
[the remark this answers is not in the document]
EthanObadia: This version is outdated, what do you think of the current one:
```python3
 rho_evol = torch.stack(
            [
                operator_product(
                    kraus_unitary, operator_product(state, kraus_dagger, self.target), self.target
                )
                for kraus_unitary, kraus_dagger in zip(self.unitary(values), self.dagger(values))
            ],
            dim=0,
        )
        rho_evol = torch.sum(rho_evol, dim=0)
        return DensityMatrix(rho_evol)

**Comment by EthanObadia** [comment]
Thanks @dominikandreasseitz for your review.
> Hi @EthanObadia, can you add some tests + some example in the docs with plots where you use the noise gate

@gvelikova suggested me ideas for `Flip` gate's tests, I will add them and continue to think about tests for most complex noise gates.

File: pyqtorch/noise.py (hunk 4 above)
Comment (EthanObadia) [reply pyqtorch/noise.py:24]: As for now, the kraus operators are hard coding by us, thus I know they are defined on the same device. You right, this test is useless. 
Removed.

File: pyqtorch/noise.py [hunk 25]
```diff
@@ -0,0 +1,368 @@
...
+    ):
+        if rate > 1.0 or rate < 0.0:
+            raise ValueError("The damping rate is not a correct probability")
+        K0: Tensor = torch.tensor([[1, 0], [0, sqrt(1 - rate)]], dtype=DEFAULT_MATRIX_DTYPE)
+        K1: Tensor = torch.tensor([[0, 0], [0, sqrt(rate)]], dtype=DEFAULT_MATRIX_DTYPE)
+        Kraus_GeneralizeAmplitudeDamping: list[Tensor] = [K0, K1]
+        super().__init__(Kraus_GeneralizeAmplitudeDamping, target, rate)
+
+
+class GeneralizeAmplitudeDamping(Noise):
```
Roland-djee, line 315: Not sure about this one. Feel free to ignore.

```suggestion
class GeneralizedAmplitudeDamping(Noise):
```

pyqtorch/utils.py
```diff
@@ -125,6 +125,10 @@ def param_dict(keys: Sequence[str], values: Sequence[Tensor]) -> dict[str, Tenso
     return {key: val for key, val in zip(keys, values)}
 
 
+class DensityMatrix(Tensor):
```
Roland-djee, line 128: Is it necessary to define a custom type if it does pretty much nothing ?

tests/test_digital.py
```diff
@@ -342,3 +350,65 @@ def test_operator_product(gate: Primitive) -> None:
     assert torch.allclose(
         op_mul, torch.eye(2**n_qubits, dtype=torch.cdouble).unsqueeze(2).repeat(1, 1, max_batch)
     )
+
+
+@pytest.mark.parametrize("FlipGate", [pyq.BitFlip, pyq.PhaseFlip, pyq.Depolarizing])
```
Roland-djee, line 355:
```suggestion
@pytest.mark.parametrize("noisy_flip_gates", [BitFlip, PhaseFlip, Depolarizing])
```

And import noise gates.

Roland-djee (commented):
Nice, it's getting in shape certainly. Thanks for this. Few comments from my side.

pyqtorch/noise.py
```diff
@@ -0,0 +1,368 @@
...
+
+
+class Noise(torch.nn.Module):
+    def __init__(
+        self, kraus: list[Tensor], target: int, probabilities: tuple[float, ...] | float
+    ) -> None:
+        super().__init__()
+        self.target: int = target
+        self.qubit_support: tuple[int, ...] = (self.target,)
+        self.kraus = []
```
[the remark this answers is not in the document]
EthanObadia: Changed.

pyqtorch/noise.py
```diff
@@ -0,0 +1,368 @@
...
+    def prob(self) -> tuple[float, ...] | float:
+        return self.probabilities
+
+    @property
+    def name(self) -> str:
+        return self.__class__.__name__
+
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, type(self)):
+            return self.name == other.name and self.probabilty == other.probabilty
```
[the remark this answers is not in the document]
EthanObadia: You're right, changed.

pyqtorch/noise.py
```diff
@@ -0,0 +1,368 @@
...
+        return False
+
+    def extra_repr(self) -> str:
+        return f"'qubit_support'={self.qubit_support}, 'probabilities'={self.probabilities}"
+
+    @property
+    def kraus_operator(self) -> list[Tensor]:
+        return self.kraus
+
+    # Since PyQ expects tensor.Size = [2**n_qubits, 2**n_qubits,batch_size].
```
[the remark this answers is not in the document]
EthanObadia: Moved.

pyqtorch/noise.py
```diff
@@ -0,0 +1,368 @@
...
+            [
+                operator_product(
+                    kraus_unitary, operator_product(state, kraus_dagger, self.target), self.target
+                )
+                for kraus_unitary, kraus_dagger in zip(self.unitary(values), self.dagger(values))
+            ],
+            dim=0,
+        )
+        rho_evol = torch.sum(rho_evol, dim=0)
+        return DensityMatrix(rho_evol)
```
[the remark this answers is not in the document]
EthanObadia: You're right. If the input state isn't a `DensityMatrix`, it wil be converted before the computation. Thus this is useless. Removed.

pyqtorch/noise.py
EthanObadia: Changed.

tests/test_digital.py
EthanObadia: Changed.

pyqtorch/noise.py
```diff
@@ -0,0 +1,368 @@
...
+
+    def __init__(
+        self,
+        target: int,
+        probability: float,
+    ):
+        if probability > 1.0 or probability < 0.0:
+            raise ValueError("The probability value is not a correct probability")
+        proba_tensor = torch.tensor(probability, dtype=DEFAULT_MATRIX_DTYPE)
+        K0: Tensor = torch.sqrt(1.0 - proba_tensor) * IMAT
```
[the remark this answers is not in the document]
EthanObadia: `torch.sqrt(1.0 - probability)` will not work as `torch.sqrt` accepts as input only `Tensor`. However, this version is works:
```python3
from math import sqrt
 K0: Tensor = sqrt(1.0 - probability)*IMAT
```
Does it suit you? And it is preferable to import `from math import sqrt` or `from numpy import sqrt`

pyqtorch/noise.py
```diff
@@ -0,0 +1,368 @@
...
+        self.qubit_support: tuple[int, ...] = (self.target,)
+        self.kraus = []
+        for index, tensor in enumerate(kraus):
+            self.register_buffer(f"kraus_{index}", tensor)
+            self.kraus.append(tensor)
+        self._device = self.kraus[0].device
+        self.probabilities: tuple[float, ...] | float = probabilities
+
+    @property
+    def prob(self) -> tuple[float, ...] | float:
```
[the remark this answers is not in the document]
EthanObadia: You're right, a getter is finally not really required. Removed.

pyqtorch/noise.py
```diff
@@ -0,0 +1,368 @@
...
+
+        Raises:
+            TypeError: If the input `state` or `kraus_list` is not a Tensor.
+        """
+        if not isinstance(state, Tensor):
+            raise TypeError("The input must be a Tensor")
+        if not isinstance(state, DensityMatrix):
+            state = density_mat(state)
+        rho_evol = torch.stack(
+            [
```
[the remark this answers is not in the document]
EthanObadia: Do you prefer this version ?
```python3
        rho_evols: list[Tensor] = []
        for kraus_unitary, kraus_dagger in zip(self.unitary(values), self.dagger(values)):
            rho_evol: Tensor = operator_product(
                kraus_unitary, operator_product(state, kraus_dagger, self.target), self.target
            )
            rho_evols.append(rho_evol)
        rho_final: Tensor = torch.stack(rho_evols, dim=0)
        rho_final = torch.sum(rho_final, dim=0)
        return rho_final
```

pyqtorch/utils.py
EthanObadia: The `DensityMatrix` class is used as a "flag" to ensure the condition in the different `forward` functions. :
```python3
if not isinstance(state, DensityMatrix):
```
It allows to check if the input state is a vector state or density matrix without any condition on the `size`, `n_qubits`... So `if not isinstance(state, DensityMatrix):` just check a `bool` right?

pyqtorch/noise.py
```diff
@@ -0,0 +1,368 @@
...
+class Noise(torch.nn.Module):
+    def __init__(
+        self, kraus: list[Tensor], target: int, probabilities: tuple[float, ...] | float
+    ) -> None:
+        super().__init__()
+        self.target: int = target
+        self.qubit_support: tuple[int, ...] = (self.target,)
+        self.kraus = []
+        for index, tensor in enumerate(kraus):
+            self.register_buffer(f"kraus_{index}", tensor)
```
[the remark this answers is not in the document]
EthanObadia: Yeah it is probably overkill. What do you think @dominikandreasseitz ?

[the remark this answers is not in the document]
EthanObadia: 1. The buffers are static. 
2. Each kraus operator is used individually in the `forward` propagation to compute the sum: 
$$S(\rho) = \sum_i K_i \rho K^{\dagger}_i $$
So, @Roland-djee I think I need to keep this way.

pyqtorch/noise.py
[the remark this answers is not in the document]
EthanObadia: Selected `from math import sqrt`.

pyqtorch/utils.py
Roland-djee: Ah right. OK I see.

pyqtorch/noise.py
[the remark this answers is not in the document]
EthanObadia: Changed.

docs/index.md
```diff
@@ -113,6 +113,151 @@ expval = pyq.expectation(circ, state, {"theta": theta}, obs)
...
+noise_prob = 0.3
+AmpD = AmplitudeDamping(0,noise_prob)
+output_state = AmpD(input_state) #It's a density matrix
+pf = PhaseFlip(1,0.7)
+output_state = pf(output_state)
+```
+
+About the noisy circuit, the initialization is the same as the non-noisy ones and the output will always be a density matrix. Let’s show its used through the simulation of a realistic $X$ gate.
+
+We know that a $X$ gate flips the state of the qubit, for instance $X|0\rangle = |1\rangle$. However, in practice, it's common for the target qubit to stay in its original state after applying $X$ due to the interactions between it and its environment. The possibility of failure can be represented by a Bitflip gate, which flips the state again after the application of the $X$ gate, returning it to its original state with a probability of `1 - gate_fidelity`.
```
Roland-djee, line 228:
```suggestion
We know that an $X$ gate flips the state of the qubit, for instance $X|0\rangle = |1\rangle$. In practice, it's common for the target qubit to stay in its original state after applying $X$ due to the interactions between it and its environment. The possibility of failure can be represented by a `BitFlip` gate, which flips the state again after the application of the $X$ gate, returning it to its original state with a probability `1 - gate_fidelity`.
```

docs/index.md
```diff
@@ -113,6 +113,151 @@ expval = pyq.expectation(circ, state, {"theta": theta}, obs)
...
+
+```python exec="on" source="material-block"
+import torch
+import pyqtorch as pyq
+
+import matplotlib.pyplot as plt
+
+input_state = pyq.product_state('00')
```
Roland-djee, line 236: Please import types. It makes it more readable in the end.

pyqtorch/noise.py
```diff
@@ -0,0 +1,356 @@
...
+    def __init__(
+        self,
+        target: int,
+        probability: float,
+    ):
+        if probability > 1.0 or probability < 0.0:
+            raise ValueError("The probability value is not a correct probability")
+        K0: Tensor = sqrt(1.0 - probability) * IMAT
+        K1: Tensor = sqrt(probability) * XMAT
+        Kraus_Bitflip: list[Tensor] = [K0, K1]
```
Roland-djee, line 119:
```suggestion
        kraus_bitflip: list[Tensor] = [K0, K1]
```

Roland-djee (approved):
Thanks @EthanObadia. Looking nice. I would lowercase your kraus variable names but other than that this is looking good. I didn't go in great detail about the tests though. Would probably need second looks on it.

pyqtorch/noise.py
EthanObadia: Changed all kraus variable name in lowercase.

docs/index.md
```diff
@@ -113,6 +113,151 @@ expval = pyq.expectation(circ, state, {"theta": theta}, obs)
...
+
+$$
+\rho = \sum_i p_i |\psi_i\rangle \langle \psi_i|
+$$
+
+The transformations of the density operator of an open quantum system interacting with its environment (noise) are represented by the super-operator $S: \rho \rightarrow S(\rho)$, often referred to as a quantum channel.
+Quantum channels, due to the conservation of the probability distribution, must be CPTP (Completely Positive and Trace Preserving). Any CPTP super-operator can be written in the following form:
+
+$$
+S(\rho) = \sum_i K_i \rho K^{\dagger}_i
```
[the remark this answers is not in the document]
EthanObadia: They are a possible representation of a superoperators, but there may be other sets of Kraus operators that describe the same operation. They aren't unique.

docs/index.md
EthanObadia: Done.

tests/test_circuit.py
```diff
@@ -202,3 +214,59 @@ def test_merge_different_batchsize(batch_size: int) -> None:
 def test_ham_expect_fail() -> None:
     ops = [pyq.Z(0), pyq.RY(1, "theta_1")]
     ham = pyq.Hamiltonian(ops)
+
+
+@pytest.mark.parametrize("n_qubits", torch.randint(low=2, high=5, size=(5, 1)))
+def test_noise_circ(n_qubits: int) -> None:
+    GATES: list = [X, Y, Z, I, H]
```
Roland-djee, line 221: No rotation gates ? And what about 2-qubit gates ?

tests/test_circuit.py
```diff
@@ -202,3 +214,59 @@ def test_merge_different_batchsize(batch_size: int) -> None:
...
+        PhaseDamping,
+        GeneralizedAmplitudeDamping,
+    ]
+    batch_size: int = torch.randint(low=1, high=5, size=(1,)).item()
+    input_state: Tensor = random_state(n_qubits, batch_size)
+    op1, op2 = random.choice(GATES), random.choice(GATES)
+    op1 = op1(torch.randint(0, n_qubits, (1,)).item())
+    op2 = op2(torch.randint(0, n_qubits, (1,)).item())
+    noise_1, noise_2 = random.choice(NOISE_GATES), random.choice(NOISE_GATES)
+    if noise_1 == PauliChannel:
```
Roland-djee, line 237: This test is good but it deserves to be reorganised I think. All this prep work until l. 262 should go in fixtures IMHO. You can have fixture that return a tuple of parameters needed next, _i.e_ `(n_qubits, op1, op2, noise_1, noise_2)` (possibly few more that I am missing here or they can also be other fixtures). Then all your test will do is circuit construction and a bunch of assertions which make it more readable and comprehensible.

tests/test_digital.py
```diff
@@ -391,3 +404,140 @@ def test_kron_batch() -> None:
         density_matrices.append(density_matrice)
     dm_expect = torch.cat(density_matrices, dim=2)
     assert torch.allclose(dm_out, dm_expect)
+
+
+@pytest.mark.parametrize("FlipGate", [BitFlip, PhaseFlip, Depolarizing, PauliChannel])
+def test_flip_gates(FlipGate: Noise) -> None:
```
Roland-djee, line 410: Same. I think it can be broken down in fixtures that'll help making the test more readble. We discuss that if you wish.

tests/test_digital.py
```diff
@@ -391,3 +404,140 @@ def test_kron_batch() -> None:
...
+                + py * density_mat(Y(target)(input_state))
+                + pz * density_mat(Z(target)(input_state))
+            ),
+        )
+
+
+@pytest.mark.parametrize(
+    "DampingGate", [AmplitudeDamping, PhaseDamping, GeneralizedAmplitudeDamping]
+)
+def test_damping_gates(DampingGate: Noise) -> None:
```
Roland-djee, line 503: Same.

Roland-djee (commented):
Thanks for this @EthanObadia it looks nice. The main point for me is to break the tests. But let's discuss it if you want more details on how to proceed.

docs/index.md
```diff
@@ -213,6 +362,7 @@ for _ in range(N_EPOCHS):
 with torch.no_grad():
     y_final = exp_fn(params, {'x': x})
 
+plt.figure()
```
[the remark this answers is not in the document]
EthanObadia: In my opinion, printing plots directly on the doc is more interesting than imposing copying and executing the code. I find it easier to understand that way. The work already done by @dominikandreasseitz prompted me to take this initiative.

In this example that was written by @dominikandreasseitz, I just added `plt.figure()` because the fact of not delimited the figures created errors to display them.

pyqtorch/utils.py
```diff
@@ -231,6 +235,41 @@ def promote_operator(operator: Tensor, target: int, n_qubits: int) -> Tensor:
...
+
+    Returns:
+        DensityMatrix: The density matrix after applying random promotion.
+
+    Raises:
+        ValueError: If `target` is not within the valid range [0, n_qubits - 1].
+    """
+    if target > n_qubits - 1:
+        raise ValueError("The target must be a valid qubit index within the circuit's range.")
+    torch.manual_seed(
```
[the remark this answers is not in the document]
EthanObadia: I would prefer to leave it here because I want this function to be general. I believe it is quite useful and could be used in other cases beyond the tests where I have used it so far. That's why I wrote it in `utils.py` instead of `tests`.

pyqtorch/utils.py
```diff
@@ -153,6 +153,10 @@ def batch_last(operator: Tensor) -> Tensor:
     return torch.permute(operator, undo_perm)
 
 
+class DensityMatrix(Tensor):
```
dominikandreasseitz, line 156: @EthanObadia whats missing here?

EthanObadia: Nothing. The `DensityMatrix` class is used as a "flag" to ensure the condition in the different `forward` functions. :
```python3
if not isinstance(state, DensityMatrix):
```
It allows to check if the input state is a vector state or a density matrix without any condition on the `size`, `n_qubits`... So `if not isinstance(state, DensityMatrix):` just check a `bool` right?

pyqtorch/primitive.py
```diff
@@ -29,9 +29,18 @@ def unitary(self, values: dict[str, Tensor] | Tensor = dict()) -> Tensor:
         return self.pauli.unsqueeze(2)
 
     def forward(self, state: Tensor, values: dict[str, Tensor] | Tensor = dict()) -> Tensor:
-        return apply_operator(
-            state, self.unitary(values), self.qubit_support, len(state.size()) - 1
-        )
+        if isinstance(state, DensityMatrix):
```
dominikandreasseitz, line 32: maybe @awennersteen has some thoughts about how to handle this

pyqtorch/utils.py
dominikandreasseitz: does this work?

EthanObadia: Yes

pyqtorch/noise.py
awennersteen: Is this even going to be differentiable?
If not, I'd say turn off the autodiff features and save a bunch of memory


Also, further to Dominiks comment.
I think (as a followup perhaps) it could be worthwhile to refactor so that we have a class that combines some of the common setup between the primitives class and this one

tests/test_circuit.py
awennersteen: very neat

pyqtorch/utils.py
```diff
@@ -153,6 +153,10 @@ def batch_last(operator: Tensor) -> Tensor:
...
+    pass
+
+
 def density_mat(state: Tensor) -> Tensor:
```
awennersteen, line 160:
```suggestion
def density_mat(state: Tensor) -> DensityMatrix:
```

pyqtorch/utils.py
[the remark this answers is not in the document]
awennersteen: I don't think it's good practise to hide a seed set like this as that is a global function, i.e. it's not limited to the scope of the function that you set it in and will impact the rest of the program.

awennersteen (commented):
nice!

tests/test_circuit.py
EthanObadia: Added rotation gates.

tests/test_circuit.py
EthanObadia: Modified.

tests/test_digital.py
EthanObadia: Changed.

tests/test_digital.py
EthanObadia: Changed

tests/test_circuit.py
EthanObadia: There is a bug with 2-qubits gates right now. So on a futur version. I'll add 2-qubit gates tests.

pyqtorch/noise.py
EthanObadia: Thank you for all your insightful comments @awennersteen.

Even though the noise gates do not have parameters to optimize directly, they need to be differentiable to allow gradient propagation throughout the circuit. This ensures that the parameters of other parametric gates, like the `RX` gate can be properly optimized. I need to ensure that all operations in `Noise` class are compatible with automatic differentiation to enable consistent and efficient optimization of the entire circuit.


Also, It would indeed be beneficial to create a parent class for both the `Primitive` class and `Noise` class to consolidate common setups. I will consider implementing this in a future PR. Thank you for the idea.

pyqtorch/noise.py
```diff
@@ -0,0 +1,356 @@
...
+        Returns:
+            Tensor: Quantum state as a density matrix after evolution.
+
+        Raises:
+            TypeError: If the input `state` or `kraus_list` is not a Tensor.
+        """
+        if not isinstance(state, Tensor):
+            raise TypeError("The input must be a Tensor")
+        if not isinstance(state, DensityMatrix):
+            state = density_mat(state)
```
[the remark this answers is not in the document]
EthanObadia: I understand the importance of maintaining interface consistency. However, the state can initially be a state vector, and it is the introduction of `Noise` that transforms it into a density matrix.
For instance, consider a circuit where gates from the `Primitive` are applied, resulting in a state vector. Upon encountering the first `Noise` gate, this state vector needs to be converted into a density matrix to proceed with the calculation. 
Thus, I believe that it is necessary to handle this initial transformation within the `Noise` gate to ensure the circuit functions correctly. 

What do you think about it ?

pyqtorch/noise.py
```diff
@@ -0,0 +1,356 @@
...
+        rho_evols: list[Tensor] = []
+        for kraus_unitary, kraus_dagger in zip(self.unitary(values), self.dagger(values)):
+            rho_evol: Tensor = operator_product(
+                kraus_unitary, operator_product(state, kraus_dagger, self.target), self.target
+            )
+            rho_evols.append(rho_evol)
+        rho_final: Tensor = torch.stack(rho_evols, dim=0)
+        rho_final = torch.sum(rho_final, dim=0)
+        return rho_final
```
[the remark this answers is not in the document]
EthanObadia: I agree that using a single `einsum` expression could enhance memory efficiency, and I can consider this as a future optimization task.
However, since I also need to work on the implementation of the `pyq` noise in `Qadence`, I believe it would be more practical to address this in a separate pull request focused on refactoring.

In that future pull request, I will consider your idea of using `einsum` and the idea of creating a DensityMatrix class that inherits from `Primitive`.

pyqtorch/utils.py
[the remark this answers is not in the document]
EthanObadia: I’ve revised the function to use a local random number generator as 
```python3
rng = torch.Generator().manual_seed(n_qubits)
```

pyqtorch/noise.py
```diff
@@ -0,0 +1,356 @@
...
+
+
+class Noise(torch.nn.Module):
+    def __init__(
+        self, kraus: list[Tensor], target: int, probabilities: tuple[float, ...] | float
+    ) -> None:
+        super().__init__()
+        self.target: int = target
+        self.qubit_support: tuple[int, ...] = (self.target,)
+        self.kraus: list = []
```
dominikandreasseitz, line 20: why is `self.kraus` a list? I suppose they are immutable tensors so they could be stored as buffers?

pyqtorch/noise.py
```diff
@@ -0,0 +1,356 @@
...
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, type(self)):
+            return self.name == other.name and self.probabilties == other.probabilties
+        return False
+
+    def extra_repr(self) -> str:
+        return f"'qubit_support'={self.qubit_support}, 'probabilities'={self.probabilities}"
+
+    @property
+    def kraus_operator(self) -> list[Tensor]:
```
dominikandreasseitz, line 40: why the property? we already have the `kraus` field in the class, can we just give it the same name? so,  `self.kraus_operator` OR `self.kraus` ? but not have both

pyqtorch/noise.py
```diff
@@ -0,0 +1,356 @@
...
+        Args:
+            state (Tensor): Input quantum state represented as a tensor.
+
+        Returns:
+            Tensor: Quantum state as a density matrix after evolution.
+
+        Raises:
+            TypeError: If the input `state` or `kraus_list` is not a Tensor.
+        """
+        if not isinstance(state, Tensor):
```
dominikandreasseitz, line 69: this check can be removed

EthanObadia: Removed.

pyqtorch/noise.py
```diff
@@ -0,0 +1,356 @@
...
+        self.qubit_support: tuple[int, ...] = (self.target,)
+        self.kraus: list = []
+        for index, tensor in enumerate(kraus):
+            self.register_buffer(f"kraus_{index}", tensor)
+            self.kraus.append(tensor)
+        self._device: torch.device = self.kraus[0].device
+        self.probabilities: tuple[float, ...] | float = probabilities
+
+    @property
+    def name(self) -> str:
```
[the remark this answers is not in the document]
EthanObadia: You are right. The `name` method is not necessary. I have removed it and adjusted the `__eq__` method to directly compare the class names. This simplifies the code and eliminates any redundant methods: 

```python3
def __eq__(self, other: object) -> bool:
        if isinstance(other, type(self)):
            return self.__class__.__name__ == other.__class__.__name__ and self.probabilities == other.probabilities
        return False
```

pyqtorch/noise.py
EthanObadia: Thank you for your suggestion. I registered the Kraus operators as buffers but I didn’t actually utilize them correctly in the initial implementation.
Thus, I removed `self.kraus` and I updated the `kraus_operator` property to dynamically access the registered buffers instead of relying on a list. This ensures that the tensors are correctly handled by Pytorch: 

```python3
@property
def kraus_operator(self) -> list[Tensor]:
        return [getattr(self, f'kraus_{i}') for i in range(len(self._buffers))]
```

pyqtorch/noise.py
EthanObadia: You’re right. Having both the kraus field and the `kraus_operator` property is redundant. So I simplified the code to use a single kraus property that directly accesses the saved buffers.

gvelikova:
@EthanObadia @dominikandreasseitz what is the state with this PR? Would it be merged soon or there are some modifications expected?

pyqtorch/primitive.py
```diff
@@ -57,9 +57,19 @@ def unitary(self, values: dict[str, Tensor] | Tensor = dict()) -> Tensor:
...
+            # error type int | tuple[int, ...] expected "int" but for now it hande only 1-qubit gate
+            return DensityMatrix(
```
dominikandreasseitz, line 62: i would prefer to have separate `_forward_statevec` and `_forward_densitymat` methods which you then call in the forward

dominikandreasseitz (approved):
thanks @EthanObadia . great work! approved, I will merge it today

pyqtorch/primitive.py
EthanObadia: Ok sure

Issue #163, opened 2024-04-26
[Feature] Add Quantum Dropout


Roland-djee:
This issue concerns the implementation of the first type of quantum regularisation techniques that will be made available to Qadence and PyQ users interested in using QNNs. In particular, these are the rotational, entangling, and canonical (keeps some entangling gates in place)  quantum dropout techniques described in https://arxiv.org/abs/2310.04120.

Roland-djee:
@callumfduffy @gvelikova Could you please add a bit more details in the description ? Thanks.

PR #187, opened 2024-06-12
[Feature] Quantum dropout

callumfduffy:
This is a work in progress example for quantum dropout.

Closes https://github.com/pasqal-io/pyqtorch/issues/163

TODOs:

- [x] address all comments @callumfduffy 
- [x] move contents of example.py into `docs/dropout.md` @callumfduffy 
- [x] final review by @gvelikova


dominikandreasseitz:
hi @callumfduffy, great work! can you describe the batching issue youre facing?

gvelikova:
@dominikandreasseitz I think the batching issue was resolved and the description is just a tad stale.

pyqtorch/circuit.py
```diff
@@ -122,6 +122,84 @@ def flatten(self) -> ModuleList:
...
+        if self.training:
+            state = self.dropout_fn(state, values)
+        else:
+            for op in self.operations:
+                state = op(state, values)
+        return state
+
+    def rotational_dropout(
+        self, state: State = None, values: dict[str, Tensor] | ParameterDict = {}
+    ) -> State:
```
dominikandreasseitz, line 149: can we add some docstrings short description of what the method is doing?

pyqtorch/utils.py
```diff
@@ -64,6 +64,22 @@ class DiffMode(StrEnum):
     ADJOINT = "adjoint"
 
 
+class DropoutMode(StrEnum):
+    """
+    Which Dropout mode to use.
+
+    Options: none          - No dropout is used.
```
dominikandreasseitz, line 71: you can remove "none"

pyqtorch/circuit.py
```diff
@@ -122,6 +122,84 @@ def flatten(self) -> ModuleList:
...
+        dropout_mode: DropoutMode = DropoutMode.ROTATIONAL,
+        dropout_prob: float = 1.0,
+    ):
+        super().__init__(n_qubits, operations)
+        self.dropout_mode = dropout_mode
+        self.dropout_prob = dropout_prob
+
+        self.dropout_fn = getattr(self, dropout_mode)
+
+    def forward(self, state: State, values: dict[str, Tensor] | ParameterDict = {}) -> State:
```
dominikandreasseitz, line 139: we dont need the if...elif. we are always using dropout now right?

dominikandreasseitz (changes requested):
hey @callumfduffy nice work, i put some comments. can you also add some documentation under `docs/index.md` (just add another section) with some more details like links to papers, short description of the used algorithms etc.

docs/index.md
```diff
@@ -379,6 +379,367 @@ def fig_to_html(fig: Figure) -> str:  # markdown-exec: hide
 print(fig_to_html(plt.gcf())) # markdown-exec: hide
 ```
 
+## Fitting a noisy sinusoid with quantum dropout
+
```
dominikandreasseitz, line 383: hey @callumfduffy can you put this into `docs/dropout.md`  (rebase on main if you havent)

pyqtorch/circuit.py
callumfduffy: kept in case you want to use the circuit at test time and do not want dropout present

docs/dropout.md
```diff
@@ -1 +1,361 @@
...
+
+import pyqtorch as pyq
+from pyqtorch.circuit import DropoutQuantumCircuit
+from pyqtorch.parametric import Parametric
+
+manual_seed(12345)
+np.random.seed(12345)
+
+
+class MinMaxScaler:
```
gvelikova, line 25: Ok, for all of the  functions and classes here I would suggest to add docstrings and comments (simpler functions may have just a sentence of what they are), but this way their functionalities will be clearer to the user.

gvelikova: Also, for this documentation treat it as a proper tutorial, which implies that you need to add more context on each of the steps taken. Check the tutorials in qadence.

pyqtorch/circuit.py
```diff
@@ -215,6 +215,112 @@ def sample(probs: Tensor) -> Counter:
             return list(map(lambda p: sample(p), probs))
 
 
+class DropoutQuantumCircuit(QuantumCircuit):
+    def __init__(
```
gvelikova, line 219: Again as stated above, add docstrings.

pyqtorch/circuit.py
```diff
@@ -215,6 +215,112 @@ def sample(probs: Tensor) -> Counter:
...
+        self,
+        n_qubits: int,
+        operations: list[Module],
+        dropout_mode: DropoutMode = DropoutMode.ROTATIONAL,  # type: ignore
```
gvelikova, line 223: Ideally, after looking into the StringEnum you would be able to unmute mypy type checks.

pyqtorch/circuit.py
```diff
@@ -215,6 +215,112 @@ def sample(probs: Tensor) -> Counter:
...
+                    keep = 1 - bernoulli(tensor(self.dropout_prob))  # type: ignore
+                    if keep:
+                        state = op(state, values)
+
+                else:
+                    state = op(state, values)
+            else:
+                state = op(state, values)
+
+        return state
```
gvelikova, line 266:
``` # mild suggestion if it works
for op in self.operations:
    if hasattr(op, "param_name") and values[op.param_name].requires_grad and 1 - bernoulli(tensor(self.dropout_prob)):
        return state
    else:
        return op(state, values) # new state
```

gvelikova: Potentially, if useful this can be applied to the code of the other types of dropout.

pyqtorch/utils.py
```diff
@@ -87,6 +87,21 @@ class DiffMode(StrEnum):
     ADJOINT = "adjoint"
 
 
+class DropoutMode(StrEnum):
+    """
+    Which Dropout mode to use.
```
gvelikova, line 92: Maybe expand on this.

dominikandreasseitz (changes requested):
Hi @callumfduffy. Thanks a lot! Some remarks:

- [x] Regarding your tutorial, Please have a look at https://github.com/pasqal-io/pyqtorch/blob/main/docs/fitting_a_function.md, i use a `hea` fixture there + maybe you can steal some of the logic to create your circuit from the circuit there too
- [x] In the same tutorial, you will find some lines at the bottom which display your figures, please just copy them into your tutorial to display your results
- [x] Please remove all the type ignores in your code if possible
- [x] Please reduce the number of epochs significantly so the docs build faster. Also raise the batch_size to 500 to you get a better fit

dominikandreasseitz:
hi @callumfduffy , @Roland-djee will take over the review here, thanks!

docs/dropout.md
```diff
@@ -1 +1,402 @@
...
+# the two feature maps we will be using
+def fm1(n_qubits):
+    return [pyq.RY(i, "x1") for i in range(n_qubits)]
+
+
+def fm2(n_qubits):
+    return [pyq.RZ(i, "x2") for i in range(n_qubits)]
+
+# The class which constructs QNNs
+class QuantumModelBase(torch.nn.Module):
```
Roland-djee, line 176: I'm a bit surprised there is no datastructure you could use here instead of reconstructing one.

pyqtorch/circuit.py
```diff
@@ -215,6 +215,114 @@ def sample(probs: Tensor) -> Counter:
...
+        for op in self.operations:
+            if (
+                (hasattr(op, "param_name"))
+                and (values[op.param_name].requires_grad)
+                and not (int(1 - bernoulli(tensor(self.dropout_prob))))
+            ):
+                continue
+
+            else:
+                state = op(state, values)
```
Roland-djee, line 269: I think I would simplify it by doing something like:

```suggestion
            has_param_name = hasattr(op, "param_name")
            requires_grad = values[op.param_name].requires_grad
            not_<name> = not (int(1 - bernoulli(tensor(self.dropout_prob)))
            if not (has_param_name and requires_grad not_<name>):
                state = op(state, values)
```

pyqtorch/circuit.py
```diff
@@ -215,6 +215,114 @@ def sample(probs: Tensor) -> Counter:
...
+        Returns:
+            State: pure state vector
+        """
+        for op in self.operations:
+            if not (hasattr(op, "param_name")) and not (
+                int(1 - bernoulli(tensor(self.dropout_prob)))
+            ):
+                continue
+            else:
+                state = op(state, values)
```
Roland-djee, line 290: Same. I would define a bunch of predicates and remove the useless `continue` clause.

pyqtorch/circuit.py
```diff
@@ -215,6 +215,114 @@ def sample(probs: Tensor) -> Counter:
             return list(map(lambda p: sample(p), probs))
 
 
+class DropoutQuantumCircuit(QuantumCircuit):
```
Roland-djee, line 218: Is there a need for another class or could this be integrated in `QuantumCircuit` ?

Roland-djee (commented):
Hey thanks @callumfduffy. Comments from my side.

pyqtorch/circuit.py
callumfduffy: unfortunately I don't believe we can predicates beforehand since "has_param_name = hasattr(op, "param_name")" will return an error for gates which do not have the param_name attribute such as the CNOT gate. The if statement avoids this issue since the statements are evaluated sequentially. However, I can re-write this so we no longer have the continue statement and just one if statement.

pyqtorch/circuit.py
callumfduffy: This was a suggestion from @dominikandreasseitz, so that the QuantumCircuit class remained not overly complex.

jpmoutinho: I agree with designing this on a separate class. All methods here are too specific to put in the QuantumCircuit.

docs/dropout.md
jpmoutinho: No, there isn't. So far there wasn't really a need for one. 

It seems to me like the usage here is just to follow the typical torch neural network construction, which in a way is a more "torchy" use of PyQ than what Qadence actually does. 

For now, if it's just for convenience of the tutorial that's fine. We can then see if a general version of this deserves a place in PyQ.

docs/dropout.md
```diff
@@ -1 +1,402 @@
...
+
+        self.embedding1 = fm1(n_qubits=n_qubits)
+        self.embedding2 = fm2(n_qubits=n_qubits)
+
+        self.params = torch.nn.ParameterDict()
+        operations = self.build_operations()
+        self.circuit = self.build_circuit(operations)
+        self.observable = pyq.QuantumCircuit(n_qubits, [pyq.Z(i) for i in range(n_qubits)]).to(
+            device=device, dtype=torch.complex64
+        )
```
jpmoutinho, line 191: Please use the `Observable` class to build observables (it's recent).

tests/test_dropout.py
```diff
@@ -0,0 +1,49 @@
...
+        dropout_prob=1.0,
+    )
+    obs = pyq.QuantumCircuit(n_qubits=n_qubits, operations=[pyq.Z(1)])
+
+    exp_circ = pyq.expectation(circuit=circ, state=state, values=values, observable=obs)
+    exp_dropout_circ = pyq.expectation(
+        circuit=dropout_circ, state=state, values=values, observable=obs
+    )
+
+    assert torch.allclose(exp_circ, exp_dropout_circ, atol=ATOL)
```
jpmoutinho, line 49: It is not clear to me what the logic of this test is. If one circuit does dropout and the other doesn't, shouldn't we expect the expectation value to be different?

If we are testing for equality in the expectation value, is it that the dropout configuration is set to never dropout? If yes, than I understand that is a good sanity check to include, but I feel like something more should be tested regarding the actual effect of the dropout methods.

But maybe I am missing something so feel free to correct me 😅

jpmoutinho (commented):
Just a quick look from my side, thanks for the work @callumfduffy :). Overall the code looks very nice and clean.

tests/test_dropout.py
callumfduffy: Sorry for the confusion I've added a docstring to the test to clear up its purpose. The tests set the dropout probability to 1 such that every gate the specific dropout method could effect is dropped, this provides a deterministic result which we can manually check against.

jpmoutinho: Ah got it, I see now that you are testing, thanks.

gvelikova (approved):

PR #191, opened 2024-06-18
[Feature] Add time-dependent Schrodinger and Lindblad equation solvers

vytautas-a:
Proposal to add time-dependent Schrodinger and Lindblad equation solvers.


vytautas-a:
Here the time-dependent solvers are simply put in a separate module, but maybe some other structure would work better. One more caveat here is that the time-dependent solver tests require `qutip` - that would be an additional dependency.

pyproject.toml
```diff
@@ -29,7 +29,7 @@ classifiers=[
     "Programming Language :: Python :: Implementation :: CPython",
     "Programming Language :: Python :: Implementation :: PyPy",
 ]
-dependencies = ["torch", "numpy"]
+dependencies = ["torch", "numpy", "qutip~=4.7.5"]
```
dominikandreasseitz, line 32: lets make qutip optional

jpmoutinho (commented):
Nice @vytautas-a, looking clean 👍 

Currently the solvers accept `H` as some generic function that takes a time value and returns `H(t)` as a dense matrix? Or sparse? Or any?

I wonder if in the future it's also worth making these more integrated with the PyQ blocks so that Hamiltonians are applied "block by block", but we would have to check if there is some efficiency to be gained there.

pyqtorch/time_dependent/integrators/adaptive.py
```diff
@@ -0,0 +1,190 @@
...
+
+        See Equation (4.12) and (4.13) of [1] for the detailed steps.
+        """
+        if error == 0:  # no error -> maximally increase the time step
+            return dt * self.options.max_factor
+
+        elif error <= 1:  # time step accepted -> take next time step at least as large
+            return float(
+                dt
+                * max(
```
Roland-djee, line 156: Weird indentation.

pyqtorch/time_dependent/integrators/krylov.py
```diff
@@ -0,0 +1,43 @@
...
+from typing import Callable
+
+import torch
+from torch import Tensor
+
+from pyqtorch.time_dependent.options import KrylovSolverOptions
+
+
+class KrylovIntegrator:
+    """Uses Krylov subspace method to evolve the state vector."""
```
Roland-djee, line 13:
```suggestion
    """Krylov subspace method to evolve the state vector."""
```

pyqtorch/time_dependent/methods/krylov.py
```diff
@@ -0,0 +1,87 @@
+from __future__ import annotations
+
+import torch
+from torch import Tensor
+
+from pyqtorch.time_dependent.integrators.krylov import KrylovIntegrator
+
+
+class Krylov(KrylovIntegrator):
```
Roland-djee, line 9: Does it deserve a base class since it is used only once ? Or is it possible to abstract `Integrator` even more to be use by all solvers ?

pyqtorch/time_dependent/solvers.py
```diff
@@ -0,0 +1,80 @@
...
+    def __init__(
+        self,
+        H: Tensor | Callable[..., Any],
+        psi0: Tensor,
+        tsave: Tensor,
+        options: AdaptiveSolverOptions,
+    ):
+        super().__init__(H, psi0, tsave, options)
+
+    def odefun(self, t: float, psi: Tensor) -> Tensor:
```
Roland-djee, line 28:
```suggestion
    def ode_fun(self, t: float, psi: Tensor) -> Tensor:
```

pyqtorch/time_dependent/solvers.py
```diff
@@ -0,0 +1,80 @@
...
+        """Compute dpsi / dt = -1j * H(psi) at time t."""
+        with warnings.catch_warnings():
+            # filter-out UserWarning about "Sparse CSR tensor support is in beta state"
+            warnings.filterwarnings("ignore", category=UserWarning)
```
Roland-djee, line 32: I think we should use the new logging infra @dominikandreasseitz ?

pyqtorch/time_dependent/solvers.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def odefun(self, t: float, rho: Tensor) -> Tensor:
+        H = self.H(t)
+        lindblad_term = sum(L @ rho @ L.mH for L in self.L_tuple)
+        out: Tensor = -1j * self.Hnh(H) @ rho + 0.5 * lindblad_term
+        return out + out.mH
+
+
+class SEDormandPrince5(SESolver, DormandPrince5):
+    pass
```
Roland-djee, line 72: To be implemented later ?

Roland-djee (approved):
Thanks @vytautas-a looking good. Not much to say, minor cosmetic comments. Generally, I think it best to add more docstrings and comment to elucidate some obscure parts.

pyqtorch/time_dependent/integrators/adaptive.py
```diff
@@ -0,0 +1,190 @@
...
+        the same notations as in the book.
+        """
+
+        sc = self.options.atol + torch.abs(y0) * self.options.rtol
+        d0, d1 = (
+            hairer_norm(y0 / sc).max().item(),
+            hairer_norm(f0.to_dense() / sc).max().item(),
+        )
+
+        if d0 < 1e-5 or d1 < 1e-5:
```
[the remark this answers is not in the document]
vytautas-a: Actually, I don't really know - I guess it's following the book reference

pyqtorch/time_dependent/methods/krylov.py
vytautas-a: Krylov integrator is a bit different than the adaptive one so I doubt that we'll be able to abstract that. However, I believe that we can come up with master equation solver that uses Krylov method. Thus, I think this class has to stay.

pyqtorch/time_dependent/solvers.py
vytautas-a: This is actually a strange part, but I couldn't think how to implement it in other way (original dynamiqs implementation follows the same path). The problem here is that I want to use two more abstract classes to construct DP5 solver specifically for Schrodinger equation. By inheriting from `MESolver` and `DormandPrince5` I get the master equation solver. If we implement in the future some other method `FooBar` then the Schrodinger equation adaptive solver will be constructed as `SEFooBar(SESolver, FooBar)`.

pyqtorch/time_dependent/integrators/adaptive.py
```diff
@@ -0,0 +1,190 @@
...
+from pyqtorch.utils import hairer_norm
+
+
+class AdaptiveIntegrator:
+    """Adaptive step-size ODE integrator.
+
+    For details about the integration method, see Chapter II.4 of [1].
+
+    [1] Hairer et al., Solving Ordinary Differential Equations I (1993), Springer
+        Series in Computational Mathematics.
```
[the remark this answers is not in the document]
vytautas-a: This class is not user-facing - all these arguments are explained in docstrings of `sesolve` and `mesolve` functions.

pyqtorch/time_dependent/integrators/adaptive.py
vytautas-a: That's the outcome of `black` formatter.

pyqtorch/time_dependent/solvers.py
vytautas-a: you mean to log this warning?

PR #194, opened 2024-06-19
[Documentation] Restructure docs in files

dominikandreasseitz:
Docs have grown significantly lately, so we organize them in separate files and keep the "pyqtorch in a nutshell" short


chMoussa (approved):
LGTM!

awennersteen (approved):
lgtm, assuming I don't need to thoroughly re-read the text


My only thought would be to add to the very top of index.md

```
PyQTorch can be used standalone as shown in these docs, or through our framework for quantum programming: (Qadence)[https:://github.com/pasqal-io/qadence].
```

Issue #192, opened 2024-06-18
[Refactor] Observable inheriting from Add, change forward and add expectation methods


chMoussa:
As mentioned in #174 , `Observable` should inherit from `Add`, and it should have an `expectation` method. The `forward` should just return the state with the same logic of `Add`, and the `expectation` method should do what the current `forward` does with the `inner_prod`.

Issue #175, opened 2024-05-16
[Refactoring] Avoid calling block to tensor on diagonal Hamiltonians


jpmoutinho:
Instead of doing block to tensor -> dense matrix -> diagonal -> we should just use block to diagonal.

mlahariya:
#282

chMoussa:
Closing with #312

PR #201, opened 2024-06-26
[Refactor, Breaking] Changes in Observable and Scale

chMoussa:
Related to issue #192.

To complete after https://github.com/pasqal-io/pyqtorch/pull/220

- [x] Rename run and forward for forward and expectation in observables.
- [x] More streamlined Scale class.
- [ ] More tests.

### DIFF
--- a/pyqtorch/analog.py
+++ b/pyqtorch/analog.py
@@ -68,7 +68,7 @@ class GeneratorType(StrEnum):
 
 class Scale(Sequence):
     """
-    Generic container for multiplying an 'Primitive', 'Sequence' or 'Add' instance by a parameter.
+    Generic container for multiplying a 'Primitive', 'Sequence' or 'Add' instance by a parameter.
 
     Attributes:
         operations: Operations as a Sequence, Add, or a single Primitive operation.
--- a/pyqtorch/analog.py
+++ b/pyqtorch/analog.py
@@ -269,7 +269,7 @@ class Add(Sequence):
         )
 
 
-class Observable(Sequence):
+class Observable(Add):
     """
     The Observable :math:`O` represents an operator from which
     we can extract expectation values from quantum states.
@@ -292,7 +292,9 @@ class Observable(Sequence):
             n_qubits = max(self.qubit_support) + 1
         self.n_qubits = n_qubits
 
-    def run(self, state: Tensor, values: dict[str, Tensor]) -> State:
+    def forward(
+        self, state: Tensor, values: dict[str, Tensor] | ParameterDict = dict()
+    ) -> State:
         """
         Apply the observable onto a state to obtain :math:`\\|O\\ket\\rangle`.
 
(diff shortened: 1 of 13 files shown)


jpmoutinho (dismissed):
Thanks @chMoussa 👍

This is a small thing but now is the time to iron out these conceptual details.

Let's wait for @dominikandreasseitz approval btw. Unfortunately this is a breaking change, so we should think carefully about the issues upstream.

pyqtorch/analog.py
```diff
@@ -359,7 +361,9 @@ def __init__(
         self.qubit_support = operations.qubit_support
         self.n_qubits = n_qubits
 
-    def run(self, state: Tensor, values: dict[str, Tensor]) -> Tensor:
+    def forward(
+        self, state: Tensor, values: dict[str, Tensor] | ParameterDict = dict()
```
dominikandreasseitz, line 365: its a bit strange to me that the forward of an  Add block does not add but just do the normal sequence forward

jpmoutinho: Sorry, you are right. The point was actually that the `forward` of the observable should be the same as the `forward` from `Add`, so actually there should no be `forward` rewrite on the `Observable` because now it inherits from `Add`.

The `DiagonalObservable` currently inherits from `Primitive` so it should be handled differently.

chMoussa: I see what you mean. Then it is better to remove the forward method to make use the one from Add.

chMoussa: Now this should be handled @dominikandreasseitz @jpmoutinho.

pyqtorch/analog.py
```diff
@@ -355,11 +335,17 @@ def __init__(
...
-        self.qubit_support = operations.qubit_support
-        self.n_qubits = n_qubits
 
-    def run(self, state: Tensor, values: dict[str, Tensor]) -> Tensor:
+        # to make an observable with only one operation
+        # we represent this by a single Primitive
+        # with operator as the input tensor
+        # to be accessed later via self.operations[0].pauli
+        # and this is compatible with the to method for conversions
+        super().__init__(n_qubits, Primitive(operator, operations.qubit_support))
```
jpmoutinho, line 344: This is ok for now to replicate the previous logic but this class will need to be reworked, essentially so we can deal with https://github.com/pasqal-io/pyqtorch/issues/175

I also don't think we need a special `DiagonalObservable` class, just a property that keeps track of the diagonalily of the individual blocks when they are composed of `Z`, `I` or `N` and a version of `apply_operator` specific for diagonal operations. This can be used throughout PyQ for applying diagonal evolutions, and also for applying diagonal gates like `CPHASE`.

But I would not do this in this MR.

jpmoutinho (dismissed):
Thanks @chMoussa! Now I think this is ok :)

pyqtorch/analog.py
chMoussa: Sure, created issue #211 for handling this.

jpmoutinho:
Closing as the changes from this MR are now done in https://github.com/pasqal-io/pyqtorch/pull/220

PR #204, opened 2024-06-27
[Fix] Handle nested param dicts

dominikandreasseitz:
avoid inferring batch_sizes from the length of dicts as values


vytautas-a (approved):
Looks good

Issue #176, opened 2024-05-16
[Refactoring, Hamevo] Optimize block_to_tensor calls for repeated execution of HamEvo


jpmoutinho:
Original text from @dominikandreasseitz :

_Right now, we call block_to_tensor in every forward pass to get the hamiltonian, which is then exponentiated in native pyq. lets find a way to not have to call block_to_tensor but rather "fill in" parameter values in a smarter way._

Actually doing that for parametric Hamiltonians will be hard. However, we should avoid doing it for non-parametric Hamiltonians. Those should only be tensorized once and then cached for repeated evaluations.

Related to https://github.com/pasqal-io/qadence/issues/134

jpmoutinho:
Closing as one possible solution has been merged in #205. There are possibly other avenues to explore, but nothing with a concrete plan, so we can open new issues for that in the future if needed.

PR #205, opened 2024-06-27
[Refactoring, Hamevo] Optimize block_to_tensor calls for repeated execution of HamEvo

chMoussa:
Avoid recomputing the evolution operator via a simple caching system. Related to issue #176

### DIFF
--- a/pyqtorch/analog.py
+++ b/pyqtorch/analog.py
@@ -455,6 +455,8 @@ class HamiltonianEvolution(Sequence):
         and sets the logic for applying hamiltonian evolution.
         time: The evolution time :math:`t`.
         operations: List of operations.
+        cache_length: LRU cache cache_length evolution operators for given set
+                    of parameter values.
     """
 
     def __init__(
@@ -463,6 +465,7 @@ class HamiltonianEvolution(Sequence):
         time: Tensor | str,
         qubit_support: Tuple[int, ...] | None = None,
         generator_parametric: bool = False,
+        cache_length: int = 1,
     ):
         """Initializes the HamiltonianEvolution.
         Depending on the generator argument, set the type and set the right generator getter.
@@ -537,6 +540,7 @@ class HamiltonianEvolution(Sequence):
 
         # to avoid recomputing hamiltonians and evolution
         self._cache_hamiltonian_evo: dict[str, Tensor] = dict()
+        self.cache_length = cache_length
 
     @property
     def generator(self) -> ModuleList:
@@ -646,7 +650,7 @@ class HamiltonianEvolution(Sequence):
         """
 
         values_cache_key = str(OrderedDict(values))
-        if values_cache_key in self._cache_hamiltonian_evo:
+        if self.cache_length > 0 and values_cache_key in self._cache_hamiltonian_evo:
             return self._cache_hamiltonian_evo[values_cache_key]
 
         if diagonal:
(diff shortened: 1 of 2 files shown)


pyqtorch/analog.py
```diff
@@ -533,6 +535,9 @@ def __init__(
             GeneratorType.PARAMETRIC_OPERATION: self._parametric_generator,
         }
 
+        # to avoid recomputing hamiltonians and evolution
+        self._cache_hamiltonian_evo: dict[str, Tensor] = {}
```
Roland-djee, line 539:
```suggestion
        self._cache_hamiltonian_evo: dict[str, Tensor] = dict()
```

I find it less ambiguous.

pyqtorch/analog.py
```diff
@@ -640,12 +644,20 @@ def tensor(
         Raises:
             NotImplementedError for the diagonal case.
         """
+
+        values_cache_key = str(OrderedDict(values))
```
Roland-djee, line 648: Wouldn't `OrderedDict(values).keys()` work ?

tests/test_analog.py
```diff
@@ -148,6 +148,11 @@ def test_hamiltonianevolution_with_types(
     assert result.size() == (batch_size,)
     assert torch.allclose(result, target, rtol=RTOL, atol=ATOL)
 
+    # test cached
```
Roland-djee, line 151: Wouldn't it be possible to check for numbers or equivalence of some sort ? Just checking for the length of an object is a bit meaningless.

Roland-djee (commented):
Thanks @chMoussa. Comment from my side.

pyqtorch/analog.py
chMoussa: I think not as I save the operator for the values of any parameters and we can obtain a different operator with different inputs.

jpmoutinho:
Thanks @chMoussa. Did you test if you get any efficiency gains with this solution? I suspect you will potentially get some speed at the cost of memory? We should be careful with that since memory might be the bottleneck when scaling computations.

chMoussa:
> Thanks @chMoussa. Did you test if you get any efficiency gains with this solution? I suspect you will potentially get some speed at the cost of memory? We should be careful with that since memory might be the bottleneck when scaling computations.

@jpmoutinho Yes this saves some time. We can change this by limiting the cache to a few items in the dict (LRU caching system).

pyqtorch/analog.py
```diff
@@ -619,17 +628,14 @@ def forward(
...
-        # we expect the user to pass it in the `values` dict
+
+        evolved_op = self.tensor(values, None)
         return apply_operator(
             state=state,
-            operator=evolve(hamiltonian, time_evolution),
+            operator=evolved_op,
             qubits=self.qubit_support,
             n_qubits=len(state.size()) - 1,
-            batch_size=max(hamiltonian.shape[BATCH_DIM], len(time_evolution)),
```
jpmoutinho, line 632: Did you check the logic of this `max` call here that you removed?

Now I don't remember exactly why it was needed, but it had something to do with evaluating a single fixed hamiltonian for a large number of different time values.

Since you removed it, and nothing breaks, either it was not really needed or it is untested 🤔

tests/test_analog.py
```diff
@@ -142,12 +143,20 @@ def test_hamiltonianevolution_with_types(
...
     result = overlap(psi_star, psi)
     assert result.size() == (batch_size,)
     assert torch.allclose(result, target, rtol=RTOL, atol=ATOL)
 
+    # test cached
+    assert len(hamevo._cache_hamiltonian_evo) == 1
+    psi_star = hamevo(psi)
+    assert len(hamevo._cache_hamiltonian_evo) == 1
+    # check value has not changed
+    assert torch.allclose(result, target, rtol=RTOL, atol=ATOL)
```
jpmoutinho, line 158: You need to recompute `result = overlap(psi_star, psi)` for this to make sense right?

jpmoutinho (approved):
Very nice @chMoussa, thanks!

My main worry is the removal of this `max` call, it feels like it could be removing some untested edge case. But maybe this is something you checked?

pyqtorch/analog.py
chMoussa: @jpmoutinho  I could not really get the logic behind it but we are testing different batchdim and time_evo length in `test_hamiltonianevolution_with_types`, and different cases. All goes so I assumed it is necessary with the new way I implemented.

jpmoutinho: Ok cool then let's assume it is fine 😅

PR #212, opened 2024-07-02
[Fix] Overload `to` method of pyq.Scale

chMoussa:
Due to `param_name` possibly be a tensor, we should overload the `to` method for `Scale` for conversion purposes.


dominikandreasseitz (approved):
thanks for that!

PR #213, opened 2024-07-03
[Feature] Single gap GPSR

dominikandreasseitz:


chMoussa (approved):

pyqtorch/gpsr.py
```diff
@@ -0,0 +1,86 @@
...
+        for op in ctx.circuit.flatten():
+            if isinstance(op, Parametric) and isinstance(op.param_name, str):
+                spectrum = torch.linalg.eigvalsh(op.pauli).reshape(-1, 1)
+                spectral_gap = torch.unique(
+                    torch.abs(torch.tril(spectrum - spectrum.T))
+                )
+                spectral_gap = spectral_gap[spectral_gap.nonzero()]
+                assert (
+                    len(spectral_gap) == 1
+                ), "PSRExpectation only works on single_gap for now."
```
jpmoutinho, line 60: The `eigvals` can be stored in the gate itself right? No need to compute this during the backward pass?

jpmoutinho: Also, I feel like it would be best to control the applicability of `PSR` before the circuit execution, and not during the `backward` pass. E.g., each differentiation mode could have a set of supported operations, and if a circuit is called with one gate that falls outside the supported operations for the requested differentiation mode it immediately breaks. Does that make sense?

chMoussa: Hey @jpmoutinho, we actually plan to quickly add the multi gap case in a very soon MR, so this check will disappear quickly. It is mostly here so @vytautas-a and myself can pick up the basic logic for `PSR` and our tasks. I will add later the methods `eigenvalues` and `eigenvalues_generator` for all the blocks similar to `qadence` in another MR.

Roland-djee (approved):
Hey @dominikandreasseitz looking good. Please bear in mind that after the earlier discussion, it will certainly serve as a template for Horqrux's implementation (@gvelikova ). @chMoussa happy to proceed quickly for the multi-gap case.

Issue #210, opened 2024-07-01
[Bug] `tensor` method incorrectly handles control gates


EthanObadia:
`tensor` method of `Primitive` does not work correctly for control gates. Specifically, it fails to differentiate between different control and target qubit configurations. As a result, it incorrectly returns the same tensor for different control-target configurations of the same gate.

The following code demonstrates the issue:

```python3
print(torch.allclose(
    pyq.CNOT(control=0, target=1).tensor(),
    pyq.CNOT(control=1, target=0).tensor()
))
# Output: True
```

The function should return different tensors for `CNOT` gates with different control and target qubits. 
A potential fix involves ensuring the function correctly accounts for the positions of control and target qubits.

jpmoutinho:
Yes the `.tensor()` method is currently mishandling the qubit support, I've also identified a different but related problem. I'm still trying to think on the best approach to solving it.

This is also relevant to the code that you have written regarding operator - operator products and promoting operators, because if we do `tensor()` on a sequence that is essentially doing an operator - operator product with (potentially) different qubit support. This all needs to be harmonized.

Issue #211, opened 2024-07-01
[Refactor] Deprecate DiagonalObservable 


chMoussa:
Wrt #201, remove `DiagonalObservable` and use a property that keeps track of the diagonalily of the individual blocks when they are composed of `Z`, `I` or `N` and a version of `apply_operator` specific for diagonal operations. This can be used throughout PyQ for applying diagonal evolutions, and also for applying diagonal gates like `CPHASE`.

dominikandreasseitz:
great idea!

Issue #158, opened 2024-04-23
[Performance] Optimize operator - operator multiplication


jpmoutinho:
Currently different size operators in the `apply_operator` product are handled by padding the smaller one with identities. Another option is to do it by manipulating the qubit indices without padding, which is likely to be more efficient.

jpmoutinho:
Closing as already tracked in https://github.com/pasqal-io/pyqtorch/issues/221

PR #220, opened 2024-07-04
[Bug, Refactor, Testing] Fix tensor logic, change Observable and Scale, add more tests

jpmoutinho:
Now this also takes care of the stuff previously in https://github.com/pasqal-io/pyqtorch/pull/201

- [x] Finally `tensor` on all primitive gates fully working and tested for arbitrary support and expanded to arbitrary support.
- [x] `tensor` on all composite operations also fully working working and tested.
- [x] `tensor` on projectors fully working and tested.
- [x] `tensor` on `HamiltonianEvolution` fully working and tested.
- [x] Tests based on `tensor` (i.e. using `_calc_mat_vec_wavefunction`) all reviewed, refactored, and centralized in `test_tensor.py`.

## Changes to `Primitive`, `Sequence`, `Add`, `Scale`, `Observable`:
- [KINDA BREAKING] The `qubit_support` property is now always ordered, which was not the case for `Primitive`. The `tensor` method will follow this order.
- The `Scale` can now be applied to any instance of `Primitive`, `Sequence` or `Add`.
- [BREAKING] The `Observable` is now a simple extension of `Add` with an extra `expectation` method.
- [BREAKING] The `DiagonalObservable` has been removed as its current logic is much slower than the normal `Observable` logic of apply all terms individually. There will be gains to get in doing with diagonal gates but will need a full redesign in a new MR.

## Changes to the `tensor` method:
- [BREAKING] There is no longer a `n_qubits` argument.
- Instead, there is an optional `full_support`

The design of `tensor` is that doing `op.tensor()` will return the matrix where the size and rows / column ordering exactly matches `op.qubit_support`. E.g.:

```
op = CNOT(0, 1)
op.qubit_support
op.tensor()[..., 0].real
---
(0, 1)
tensor([[1., 0., 0., 0.],
        [0., 1., 0., 0.],
        [0., 0., 0., 1.],
        [0., 0., 1., 0.]], dtype=torch.float64)
```
```
op = CNOT(1, 0)
op.qubit_support
op.tensor()[..., 0].real
---
(0, 1)
tensor([[1., 0., 0., 0.],
        [0., 0., 0., 1.],
        [0., 0., 1., 0.],
        [0., 1., 0., 0.]], dtype=torch.float64)
```
```
op = CNOT(23, 12)
op.qubit_support
op.tensor()[..., 0].real
---
(12, 23)
tensor([[1., 0., 0., 0.],
        [0., 0., 0., 1.],
        [0., 0., 1., 0.],
        [0., 1., 0., 0.]], dtype=torch.float64)
```
Expanding an operator by introducing explicit identity matrices can be done with the `full_support` argument. The order in which the `full_support` tuple is given does not matter, it will simply sort it, check which qubits in it are not part of the support of the operator, and add those identities. The logic for this is in `utils.expand_operator` (which would eventually replace `promote_operator`)
```
op = CNOT(23, 12)
op.tensor(full_support = (2, 12, 23))[..., 0].real
---
(12, 23)
tensor([[1., 0., 0., 0., 0., 0., 0., 0.],
        [0., 0., 0., 1., 0., 0., 0., 0.],
        [0., 0., 1., 0., 0., 0., 0., 0.],
        [0., 1., 0., 0., 0., 0., 0., 0.],
        [0., 0., 0., 0., 1., 0., 0., 0.],
        [0., 0., 0., 0., 0., 0., 0., 1.],
        [0., 0., 0., 0., 0., 0., 1., 0.],
        [0., 0., 0., 0., 0., 1., 0., 0.]], dtype=torch.float64)
```
The logic for composite operations is exactly the same, it simply calls `tensor()` on all individual operations but expands their qubit support according to the qubit support of the composite operation itself. E.g.:
```
op = Sequence([X(0), CNOT(2, 5)])
op. qubit_support # (0, 2, 5)
op.tensor()[..., 0].real
---
tensor([[0., 0., 0., 0., 1., 0., 0., 0.],
        [0., 0., 0., 0., 0., 1., 0., 0.],
        [0., 0., 0., 0., 0., 0., 0., 1.],
        [0., 0., 0., 0., 0., 0., 1., 0.],
        [1., 0., 0., 0., 0., 0., 0., 0.],
        [0., 1., 0., 0., 0., 0., 0., 0.],
        [0., 0., 0., 1., 0., 0., 0., 0.],
        [0., 0., 1., 0., 0., 0., 0., 0.]], dtype=torch.float64)
```
The operation above will call the `X(0).tensor(full_support = (0, 2, 5))` and `CNOT(2, 5).tensor(full_support = (0, 2, 5))` and then combine them through matrix multiplication. In `Add`, the individual tensors are added together, and in `Scale`, the tensor is multiplied by the scale value. An even larger support could be passed onto the operation above by doing e.g. `op.tensor(full_support = (0, 2, 5, 43))`, which would then be propagated down the tree of operation tensors.

## What is left to do (in other MRs):
- Better polish the `Primitive` and `Parametric` classes: https://github.com/pasqal-io/pyqtorch/issues/240
- More efficient operator-operator products: https://github.com/pasqal-io/pyqtorch/issues/221 and https://github.com/pasqal-io/pyqtorch/issues/158, (need to update the issue) which would be important to make some of the `tensor` logic and noisy simulations more efficient by not using explicit identities.
- Introduce specific logic for diagonal operators: https://github.com/pasqal-io/pyqtorch/issues/211

## What this MR closes:
- Completes some of the things in https://github.com/pasqal-io/pyqtorch/issues/225, and some of the points there are now tracked elsewhere. Let's keep it open for now though
- Closes https://github.com/pasqal-io/pyqtorch/issues/210
- Closes https://github.com/pasqal-io/pyqtorch/issues/192


pyqtorch/bitstrings.py
```diff
@@ -0,0 +1,31 @@
...
+        for basis in basis_ordered
+    ]
+
+
+def permute_basis(mat: torch.Tensor, qs: tuple) -> torch.Tensor:
+    """
+    Takes an operator tensor and permutes the rows and
+    columns according to the order of the qubit support.
+    """
+    perm = _basis_order(qs)
```
awennersteen, line 30: nit:
```suggestion
def permute_basis(mat: torch.Tensor, qubit_support: tuple[int]) -> torch.Tensor:
    """
    Takes an operator tensor and permutes the rows and
    columns according to the order of the qubit support.
    """
    perm = _basis_order(qubit_support)
```

awennersteen (commented):
This makes a lot of sense to me, and also to put it in one MR.

pyqtorch/bitstrings.py
```diff
@@ -0,0 +1,35 @@
+from __future__ import annotations
+
+import numpy as np
+import torch
+
+
+def _basis_order(qs: tuple[int]) -> list[int]:
+    """
```
chMoussa, line 8: Maybe a `torch.permute(mat, list_indices)` with `list_indices` corresponding to the index ordering of the original qubit support ?

chMoussa (commented):
Lgtm, very nice redesign, let's keep up thinking about it! Just needs to bump the version at some point.

tests/conftest.py
```diff
@@ -32,26 +56,70 @@
...
         init_state: Initial state to apply block on.
         values: Values of parameter if block is parametric.
 
     Returns:
         Tensor: The new wavefunction after applying the block.
 
     """
-    mat = block.tensor(n_qubits=len(block.qubit_support), values=values)
+    mat = block.tensor(values=values, full_support=full_support)
+    qubits = block.qubit_support if full_support is None else full_support
```
Roland-djee, line 76:
```suggestion
    qubit_support = block.qubit_support if full_support is None else full_support
```

tests/test_differentiation.py
```diff
@@ -65,7 +65,7 @@ def test_adjoint_diff(n_qubits: int, n_layers: int) -> None:
     # )
 
 
-@pytest.mark.xfail  # investigate
+@pytest.mark.xfail  # Adjoing Scale is currently not supported
```
Roland-djee, line 68:
```suggestion
@pytest.mark.xfail  # Adjoining Scale is currently not supported
```

tests/conftest.py
```diff
@@ -32,26 +56,70 @@
...
     return apply_operator(
-        init_state, mat, qubits=block.qubit_support, n_qubits=n_qubits
+        init_state,
+        mat,
+        qubits=qubits,
     )
 
 
+def _get_op_support(op: type[Primitive] | type[Parametric], n_qubits: int) -> tuple:
```
Roland-djee, line 84: Maybe it would be useful to turn these functions into `@fixture` since they are used in tests ?

Roland-djee (commented):
Thanks a million @jpmoutinho this is a massive PR and very much needed. Comments from my side but generally I would be very wary about introducing new variables for support handling: there is `qubit_support`, `orig_support`, `full_support` and sometimes `temp_support`. I am aware that sometimes it is a necessary evil but it would certainly pay off in terms of clarity to avoid them as much as possible.

pyqtorch/utils.py
```diff
@@ -315,10 +316,36 @@ def operator_kron(op1: Tensor, op2: Tensor) -> Tensor:
...
+            "Expanding tensor operation requires a qubit support larger than original support."
+        )
+    device = operator.device
+    full_support = tuple(sorted(full_support))
+    temp_support = qubit_support
+    mat = operator
+    for i in full_support:
+        if i not in qubit_support:
+            temp_support += (i,)
+            other = IMAT.clone().to(device).unsqueeze(2)
```
chMoussa, line 337: @jpmoutinho add the dtype as well when applying `to`

pyqtorch/circuit.py
```diff
@@ -135,23 +135,33 @@ def flatten(self) -> ModuleList:
...
             raise NotImplementedError
-        if n_qubits is None:
-            n_qubits = max(self.qubit_support) + 1
+        if full_support is None:
+            full_support = self.qubit_support
+        elif not set(self.qubit_support).issubset(set(full_support)):
+            raise ValueError(
+                "Expanding tensor operation requires a qubit support larger than original support."
+            )
         mat = IMAT.clone().unsqueeze(2).to(self.device)
```
chMoussa, line 149: Add dtype here too

pyqtorch/circuit.py
```diff
@@ -135,23 +135,33 @@ def flatten(self) -> ModuleList:
...
-        for _ in range(n_qubits - 1):
+        for _ in range(len(full_support) - 1):
             mat = torch.kron(mat, IMAT.clone().unsqueeze(2).to(self.device))
```
chMoussa, line 151: Add dtype as well here

pyqtorch/bitstrings.py
jpmoutinho: Hmm possibly something like that could work. I think it would for permuting the qubits on the wavefunction, but the matrices don't have the same dimensions. But maybe there is a smart way to reshape into that, permute, and then reshape back. I will think about it.

pyqtorch/bitstrings.py
```diff
@@ -0,0 +1,35 @@
...
+
+    Helper function to permute the rows and columns of any operator.
+
+    TODO: There is probably a smarter way to write this function that
+    avoids the exponential overhead of explicitly writing down the basis states.
+    """
+    n_qubits = len(qs)
+    qs_ordered = np.argsort(qs)
+    qs_ranked = np.argsort(qs_ordered)
+    basis_strings = [format(k, "0{}b".format(n_qubits)) for k in range(2**n_qubits)]
```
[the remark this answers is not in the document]
jpmoutinho: Yes in Qadence they are

pyqtorch/bitstrings.py
jpmoutinho: I meant to do it but forgot 😅

pyqtorch/circuit.py
```diff
@@ -135,23 +135,33 @@ def flatten(self) -> ModuleList:
     def tensor(
         self,
         values: dict[str, Tensor] = dict(),
-        n_qubits: int | None = None,
+        full_support: tuple[int, ...] | None = None,
         diagonal: bool = False,
```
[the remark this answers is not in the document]
jpmoutinho: In some cases, but that argument is not being used at all for now. Still need to think how the diagonal handling should be. Maybe the diagonal gates will be handled as `diagonal = True` by default, such that `tensor()` gives back a 1-D vector, but then passing `diagonal = False` gives the full matrix.

pyqtorch/primitive.py
```diff
@@ -77,7 +76,10 @@ def unitary(
         values: dict[str, Tensor] | Tensor = dict(),
         embedding: Embedding | None = None,
     ) -> Tensor:
-        return self.pauli.unsqueeze(2) if len(self.pauli.shape) == 2 else self.pauli
+        mat = self.pauli.unsqueeze(2) if len(self.pauli.shape) == 2 else self.pauli
+        if self.orig_support != self.qubit_support:
```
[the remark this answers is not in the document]
jpmoutinho: Yeah, I added `self.orig_support` solely because I needed to make this comparison. Maybe there is a simpler way.

pyqtorch/utils.py
```diff
@@ -315,10 +316,36 @@ def operator_kron(op1: Tensor, op2: Tensor) -> Tensor:
...
+    if not set(qubit_support).issubset(set(full_support)):
+        raise ValueError(
+            "Expanding tensor operation requires a qubit support larger than original support."
+        )
+    device = operator.device
+    full_support = tuple(sorted(full_support))
+    temp_support = qubit_support
+    mat = operator
+    for i in full_support:
+        if i not in qubit_support:
```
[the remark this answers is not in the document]
jpmoutinho: Yes I think it would be equivalent

docs/dropout.md
```diff
@@ -186,7 +186,7 @@ class QuantumModelBase(torch.nn.Module):
         self.params = torch.nn.ParameterDict()
         operations = self.build_operations()
         self.circuit = self.build_circuit(operations)
-        self.observable = pyq.Observable(n_qubits, [pyq.Z(i) for i in range(n_qubits)]).to(
+        self.observable = pyq.Observable([pyq.Z(i) for i in range(n_qubits)]).to(
```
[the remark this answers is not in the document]
jpmoutinho: Yeah the `Observable` was completely changed, but indeed now it is more consistent and conceptually correct with the definition of an observable. Although I'm still uncertain if it's even worth keeping. It is only a wrapper of the `Add` composition class with one extra method for computing the expectation.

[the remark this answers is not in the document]
Roland-djee: OK I see. Overall, I think avoiding passing `n_qubits` around is a nice improvement in usability.

pyqtorch/utils.py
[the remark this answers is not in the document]
Roland-djee: What I meant is that _I think_ you could replace these two lines with a set difference.

pyqtorch/primitive.py
```diff
@@ -258,23 +253,25 @@ def __init__(self, control: int | tuple[int, ...], target: tuple[int, ...]):
...
+        self.control: tuple = (control,) if isinstance(control, int) else control
         mat = OPERATIONS_DICT[gate]
         mat = _controlled(
             unitary=mat.unsqueeze(2),
             batch_size=1,
             n_control_qubits=len(self.control),
         ).squeeze(2)
         super().__init__(mat, target)
-        self.qubit_support = self.control + (self.target,)  # type: ignore[operator]
+        self.orig_support = self.control + (self.target,)  # type: ignore [operator]
```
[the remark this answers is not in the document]
jpmoutinho: Yes but it goes into a full sequence of mypy complaining in other places, and fully fixing it implies changing a few more things of `Primitive`. It will be handled with #240.

pyqtorch/primitive.py
```diff
@@ -48,16 +44,19 @@ def __init__(
...
         )
         if isinstance(target, np.integer):
-            self.qubit_support = (target.item(),)
+            qubit_support = (target.item(),)
         self.register_buffer("pauli", pauli)
         self.pauli_generator = pauli_generator
         self._device = self.pauli.device
         self._dtype = self.pauli.dtype
 
+        self.orig_support = qubit_support
```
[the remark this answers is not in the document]
jpmoutinho: Ok I made it `self._qubit_support`. It is only needed for the comparison below, and does not need to be user facing.

tests/conftest.py
jpmoutinho: Well I wrote them more as helper functions. Maybe there would be some way to turn them more into fixtures... For now I just made a new `helpers.py` file not to bloat `conftest.py` with non-fixtures.

chMoussa (approved):

tests/test_tensor.py
```diff
@@ -0,0 +1,234 @@
+from __future__ import annotations
+
+import random
+
+import pytest
+import torch
+from helpers import calc_mat_vec_wavefunction, get_op_support, random_pauli_hamiltonian
```
Roland-djee, line 7: I still believe this is exactly what `@fixture` are for.

Roland-djee (approved):
Thanks a lot @jpmoutinho this is looking really good. Final comments from me. Feel free to have a look or ignore them.

pyqtorch/circuit.py
```diff
@@ -135,23 +135,38 @@ def flatten(self) -> ModuleList:
...
-            mat = torch.kron(mat, IMAT.clone().unsqueeze(2).to(self.device))
-
+        if full_support is None:
+            full_support = self.qubit_support
+        elif not set(self.qubit_support).issubset(set(full_support)):
+            raise ValueError(
+                "Expanding tensor operation requires a `full_support` argument "
+                "larger than or equal to the `qubit_support`."
+            )
+        mat = torch.eye(
```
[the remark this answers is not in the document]
jpmoutinho: Not as a separate function to build the matrix. There is just `IMAT` in `pyqtorch.matrices` for the 2x2 case.

pyqtorch/gpsr.py
```diff
@@ -100,7 +100,7 @@ def forward(
         ctx.embedding = embedding
         values = param_dict(param_names, param_values)
         ctx.out_state = circuit.run(state, values)
-        ctx.projected_state = observable.run(ctx.out_state, values)
+        ctx.projected_state = observable(ctx.out_state, values)
```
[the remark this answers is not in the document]
jpmoutinho: Currently only the `QuantumCircuit` has a `run()` method. All other "composite" operators follow the same interface as individual operations: the `forward` method returns the statevector after applying the operator, $|\psi\rangle = \hat{O}|\phi\rangle$. For individual operators, that means simply applying the operator itself. For composite operators, that means applying each of the "leaf" operators according to some rule: the `Sequence` applies each one in sequence, the `Add` applies each one separately and sums the outcome, the `Scale` applies the leaf operator and multiplies with the scale. The `Observable` is just a wrapper on top of `Add`, so it will behave exactly the same, with the added convenience of there being an `expectation` method that returns $\langle\phi|\hat{O}|\phi\rangle$.

pyqtorch/primitive.py
```diff
@@ -48,16 +44,19 @@ def __init__(
...
         )
         if isinstance(target, np.integer):
-            self.qubit_support = (target.item(),)
+            qubit_support = (target.item(),)
         self.register_buffer("pauli", pauli)
         self.pauli_generator = pauli_generator
         self._device = self.pauli.device
         self._dtype = self.pauli.dtype
 
+        self._qubit_support = qubit_support
```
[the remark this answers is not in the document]
jpmoutinho: I will revisit this in the `Primitive` + `Parametric` mr, added point [there](https://github.com/pasqal-io/pyqtorch/issues/240). I tried in this MR to change a bit how the subclasses handle the qubit support but there were issues related to the things that MR will need to solve anyway.

tests/test_tensor.py
jpmoutinho: Tbh I am not sure I fully understand fixtures. It's just that to me they are "fixed" things that are reused in the setup of the tests, and these are instead functions that act on some input and produce some output which is different every time.

The `random_pauli_hamiltonian` I can see how it could be made a fixture, but not the other two.

pyqtorch/bitstrings.py
jpmoutinho: Not solved yet, tracked now in https://github.com/pasqal-io/pyqtorch/issues/241

PR #223, opened 2024-07-09
[Refactor] Add eigenvals_generator for parametric operations

chMoussa:
To avoid calculating eigenvalues for several parametric operations.

### DIFF
--- a/pyqtorch/parametric.py
+++ b/pyqtorch/parametric.py
@@ -415,7 +415,9 @@ class ControlledRotationGate(Parametric):
                     device=self.device,
                     dtype=self.dtype,
                 ),
-                pauli_singleq_eigenvalues.flatten().to(device=self.device, dtype=self.dtype),
+                pauli_singleq_eigenvalues.flatten().to(
+                    device=self.device, dtype=self.dtype
+                ),
             )
         ).reshape(-1, 1)
 
@@ -538,7 +540,9 @@ class CRY(ControlledRotationGate):
                     device=self.device,
                     dtype=self.dtype,
                 ),
-                pauli_singleq_eigenvalues.flatten().to(device=self.device, dtype=self.dtype),
+                pauli_singleq_eigenvalues.flatten().to(
+                    device=self.device, dtype=self.dtype
+                ),
             )
         ).reshape(-1, 1)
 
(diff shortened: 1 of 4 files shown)


dominikandreasseitz (changes requested):
thanks @chMoussa , however i do think that we need to handle the "calculation of eigenvals on the fly" for arbitrary generators in the HamEvo case. for example: if the user passes a custom tensor as a generator, we store it in a Primitive instance. when we then do GPSR, we should calculate those eigenvals individually. tagging @jpmoutinho for this opinion

chMoussa:
@dominikandreasseitz Now I added an optional argument to handle the case of passing the generator. Tested this a bit more in tests by checking spectral gaps.

pyqtorch/parametric.py
```diff
@@ -395,6 +475,23 @@ def __init__(
...
+
+        Arguments:
+            values: Parameter values.
+
+        Returns:
+            Eigenvalues of the generator operator.
+        """
+        return torch.cat(
+            (
+                torch.zeros(2 ** len(self.qubit_support) - 2),
```
dominikandreasseitz, line 490: be very careful when instantiating torch tensors. always use the device and dtype of the gate. tagging @jpmoutinho just in case this gets lost for me

chMoussa: Gotcha @dominikandreasseitz, they are now instantiated with `self.device` and `self.dtype`

pyqtorch/parametric.py
```diff
@@ -15,6 +16,8 @@
 from pyqtorch.primitive import Primitive
 from pyqtorch.utils import Operator
 
+pauli_singleq_eigenvalues = torch.tensor([[-1.0], [1.0]], dtype=torch.cdouble)
```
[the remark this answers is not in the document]
chMoussa: Ok the `dtype` is changed now when calling `eigenvals_generator`

pyqtorch/parametric.py
```diff
@@ -335,6 +398,27 @@ def extra_repr(self) -> str:
...
+            Eigenvalues of the generator operator.
+        """
+        return torch.cat(
+            (
+                torch.zeros(
+                    2 ** len(self.qubit_support) - 2,
+                    device=self.device,
+                    dtype=self.dtype,
+                ),
+                pauli_singleq_eigenvalues.flatten(),
```
dominikandreasseitz, line 418: same here again, this will default to device 'cpu' and 'dtype' cdouble and error if the gate is on GPU and cfloat

chMoussa: Fixed by also by `pauli_singleq_eigenvalues.flatten().to(...)` @dominikandreasseitz

dominikandreasseitz (approved):
thanks @chMoussa

Issue #221, opened 2024-07-05
[Noise, Refactor] Generalize the Function OperatorProduct 


EthanObadia:
Issue highjacked by @jpmoutinho:

The goal of this issue should be to design a function that multiplies two operators acting on a different qubit support *without* explicitly padding the smaller operator with identities. This should be possible, and it should follow a similar logic to what the `apply_operator` function does. There, it multiplies an "small" `operator` on a "large" `state` by some smart reshaping and einsumming. The new function should do it for multiplying a "small" `operator` on a "large" `operator`.

This assumes that the `qubit_support` of the small operator is a subset of the `qubit_support` of the large operator. There is also the case where they only partially overlap. Then there is the "easy" case where they are disjoint and the solution is simply to kron them.

**Previous text:**
We need to modify the `operator_product` function to generalize it and make it more versatile for our needs, particularly for its use in the `tensor` method of the `Sequence` class. Currently, `operator_product` is highly focused on the multiplication between a density matrix and an operator, which limits the generality of this function. 

`operator_product` will be modified to remove the `target` attribute and instead accept the qubit supports for both operators as inputs. This means that the function logic will be change to operate based on the provided qubit supports, ensuring that it can handle a more general set of operator multiplications:
```python3
def operator_product(op1:Tensor, op1_qs: list, op2:Tensor, op2_qs: list) -> Tensor:
```

**Modification of `promote_operator`:**

In conjunction with the changes to `operator_product`, we also need to modify the `promote_operator` function. The aim is to eliminate issue #183 related to dimension mismatches, which currently hinder the creation of circuits with multiple qubit gates, especially in the presence of noise. The updated PromoteOperator will ensure that our circuits can handle multiple qubit gates effectively, even under noisy conditions.

jpmoutinho:
Thanks @EthanObadia. I would suggest that the `promote_operator` function should also use a similar signature of `(op: Tensor, op_qs: tuple, target_qs: tuple)`. So it compares the qubit support of the current operator and fills in the identities to have it match the target qubit support.

jpmoutinho:
@vytautas-a this issue was opened by @EthanObadia and it is related to what I had in mind, but I will need to add a few more details.

EthanObadia:
@jpmoutinho , I’m noting a few remarks here that might already be on your radar, related to operator multiplication in pyq.

[...]

PR #224, opened 2024-07-09
[Bug, Refactor] Modify tensor method

EthanObadia:
This pull request includes the following modifications :

**1. Bug Fix:**
- [x] The `tensor` method has been modified to correct the bug mentioned in issue #210  
- [x] It is now possible to call `tensor` for control gates, as well as for the special definition of `SWAP` and `CSWAP` gates, following the logic of the `BlockToTensor` method from qadence.

**2.Modification of the tensor Method for `operator_product`:**

- [x] Remove the `promote_operator` method because the `tensor` method now automatically handles the promotion by accepting the `n_qubits` parameter as needed. 
For example, if a gate is defined on 2 qubits (control and target) and it needs to be defined for a 4-qubit system, setting `n_qubits` to 4 is now sufficient.
- [x] Consequently, the `operator_product` function has been modified to no longer depend on `promote_operator` and we remove the `target` attribut. It now assumes that it receives tensors with the appropriate size, eliminating the need for further modifications #221.

**3.Modifications link to the forward passes of digital gates:**
- [x] Create the `apply_density_mat` to directly compute $O \rho O^{\dagger}$
- [x] Modify the `Noise` and `Primitive` `forward` methods according all the modification bellow. 

**4.Add tests:**

### DIFF
--- a/pyqtorch/primitive.py
+++ b/pyqtorch/primitive.py
@@ -7,9 +7,10 @@ from typing import Any
 
 import numpy as np
 import torch
+from numpy import log2
 from torch import Tensor
 
-from pyqtorch.apply import apply_operator, operator_product
+from pyqtorch.apply import apply_density_mat, apply_operator, operator_product
 from pyqtorch.matrices import (
     IMAT,
     OPERATIONS_DICT,
@@ -72,20 +73,14 @@ class Primitive(torch.nn.Module):
         self, state: Tensor, values: dict[str, Tensor] | Tensor = dict()
     ) -> Tensor:
         if isinstance(state, DensityMatrix):
-            # TODO: fix error type int | tuple[int, ...] expected "int"
-            # Only supports single-qubit gates
-            return DensityMatrix(
-                operator_product(
-                    self.unitary(values),
-                    operator_product(state, self.dagger(values), self.target),  # type: ignore [arg-type]
-                    self.target,  # type: ignore [arg-type]
-                )
-            )
+            n_qubits = int(log2(state.size(1)))
+            return apply_density_mat(self.tensor(values, n_qubits), state)
         else:
             return apply_operator(
                 state, self.unitary(values), self.qubit_support, len(state.size()) - 1
             )
 
+    # ? Do we need to keep this method now ?
     def dagger(self, values: dict[str, Tensor] | Tensor = dict()) -> Tensor:
         return _dagger(self.unitary(values))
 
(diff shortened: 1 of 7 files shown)


pyqtorch/primitive.py
```diff
@@ -200,9 +195,28 @@ def __init__(self, target: int):
...
+        if n_qubits < max(self.qubit_support) + 1:
+            n_qubits = max(self.qubit_support) + 1
+        seq = Sequence(
+            [
+                CNOT(control=self.control, target=self.target),  # type: ignore[arg-type]
+                CNOT(control=self.target, target=self.control),  # type: ignore[arg-type]
+                CNOT(control=self.control, target=self.target),  # type: ignore[arg-type]
+            ]
+        )
+        return seq.tensor(values, n_qubits)
```
jpmoutinho, line 219: Why is this needed, is there not a SWAP matrix stored that can be used directly?

pyqtorch/primitive.py
```diff
@@ -217,6 +231,30 @@ def __init__(self, control: int | tuple[int, ...], target: tuple[int, ...]):
...
+                Toffoli(
+                    control=(self.control[0], self.target[0]), target=self.target[1]  # type: ignore[index]
+                ),
+                Toffoli(
+                    control=(self.control[0], self.target[1]), target=self.target[0]  # type: ignore[index]
+                ),
+            ]  #! Can't take more that 1 control in this logic
+        )
+        return seq.tensor(values, n_qubits)
+
```
jpmoutinho, line 257: Same here.

pyqtorch/primitive.py
```diff
@@ -228,11 +266,45 @@ def __init__(self, gate: str, control: int | tuple[int, ...], target: int):
...
+        """
+        c_mat = (
+            I(target=self.control[0]).tensor(values, n_qubits)
+            - proj1.tensor(values, n_qubits)
+            + operator_product(
+                proj1.tensor(values, n_qubits),
+                self.gate(self.target).tensor(values, n_qubits),
+            )
+        )
+        return c_mat
```
jpmoutinho, line 306: Related to the previous comments, using the `Sequence` here feels a bit too hacky...

jpmoutinho (commented):
Thanks a lot for the work @EthanObadia! It's really nice that you managed to incorporate the `promote_operator` function, definitely starts to feel more consistent.

Currently I am still a little wary of the usage of `n_qubits` as the main building block of the logic behind the different tensor combinations. I feel like there are important cases that are not covered by that. But I understand that is the logic that is currently there, I think I need to run a few tests. Let's sync on Thursday about this :)

PR #226, opened 2024-07-12
[Refactoring] Add `noise` parameter and allow multi-qubit gates in noisy circuit

### LINKED ISSUE #200: [Proto] Noise modifications for Qadence implementation
This issue presents a prototype idea to modify noise for the `Qadence` implementation, where the noise is implemented such as: 

The goal is to create a way that facilitates the addition of noise into a circuit, specifically after a primitive block, and to optimize this process.

**Proposed Solution:**

**1. Add a Noise Parameter to Primitive Gates in `pyq`:**
	•	Introduce a noise parameter to primitive gates in `pyq`.
	•	By default, this noise parameter is set to `None`. If it remains `None`, the gate behaves as a standard primitive gate.
	•	If the noise parameter is an instance of `Noise`, the input matrix to the primitive class is modified to include noise.

For instance:
```python3
class X(Primitive):
    def __init__(self, target: int, noise: Noise | dict[str, Noise] | None = None):
        if noise:
            #Modify the input tensor
            super().__init__(X_noisy, target)
        else:
            super().__init__(X_noisy, target)
```

**2.	Matrix Modification for Noisy Gates:**
	•	For example, to implement a `X` gate with `BitFlip` noise:
	•	If the noise parameter is set, modify the matrix to incorporate the noise and **automate** it.
	•	The noisy gate matrix would be calculated as $X(\text{Kraus}_1 +\text{Kraus}_2)$, following the forward pass of a density matrix to compute: 
$$X(\text{Kraus}_1 +\text{Kraus}_2) \rho (\text{Kraus}^{\dagger}_1 +\text{Kraus}^{\dagger}_2)X^{\dagger}.$$

**Discussion Points:**
	•	What happens to the Noise class, given that we never create actual noisy gates and thus the forward pass isn’t very useful?
	•	The benefits and drawbacks of modifying primitive gate matrices directly.

I look forward to feedback and suggestions on this prototype.

### LINKED ISSUE #210: [Bug] `tensor` method incorrectly handles control gates

### LINKED ISSUE #221: [Noise, Refactor] Generalize the Function OperatorProduct

EthanObadia:
This pull request includes the following modifications :

### I.  Allow multi-qubit gate in noisy circuit by using `tensor` method

**1. Bug Fix:**
- [x] The `tensor` method has been modified to correct the bug mentioned in issue #210  
- [x] It is now possible to call `tensor` for control gates, as well as for the special definition of `SWAP` and `CSWAP` gates, following the logic of the `BlockToTensor` method from qadence. (in fact I will not use it -> can be erased because not really well implementing @jpmoutinho)

**2.Modification of the tensor Method for `operator_product`:**

- [x] Remove the `promote_operator` method because the `tensor` method now automatically handles the promotion by accepting the `n_qubits` parameter as needed. 
For example, if a gate is defined on 2 qubits (control and target) and it needs to be defined for a 4-qubit system, setting `n_qubits` to 4 is now sufficient.
- [x] Consequently, the `operator_product` function has been modified to no longer depend on `promote_operator` and we remove the `target` attribut. It now assumes that it receives tensors with the appropriate size, eliminating the need for further modifications #221.

**3.Modifications link to the forward passes of digital gates:**
- [x] Create the `apply_density_mat` to directly compute $O \rho O^{\dagger}$
- [x] Modify the `Noise` and `Primitive` `forward` methods according all the modification bellow.

**4.Add tests:**
- [x] Add test for the new function `apply_density_mat`
- [x] Modify test for the new function `test_operator_product`

### II. Add `noise` as a parameter of gates.

- [x] Create a class `Noisy_protocols` as the `Noise` class from `qadence`.
- [x]  Add a `noise` attribut to the `primitive` class. This attribut will be an instance of `Noisy_protocols`.
- [x]  Modify `operator_product` to directly compute $U \rho U^{\dagger}$ instead of $U \rho$.
- [x] Modify the `Primitive` `forward` function to call the `Noise` one.
- [x] Modify the `Noise` `forward` function
- [x] Modify the noise gates unit tests: `test_flip_gates`,  `test_damping_gates`, `test_noise_circ`.
- [x]  Add test on the primitive gates `noise` parameter
- [ ] Modify the noise part of the documentation.

Closes #210 
Closes #221 
Closes #200.

### DIFF
--- a/docs/analog.md
+++ b/docs/analog.md
@@ -1,0 +1,165 @@
+## Analog Operations
+
+An analog operation is one whose unitary is best described by the evolution of some hermitian generator, or Hamiltonian, acting on an arbitrary number of qubits. For a time-independent generator $\mathcal{H}$ and some time variable $t$, the evolution operator is $\exp(-i\mathcal{H}t)$. `pyqtorch` provides the HamiltonianEvolution class to initialize analog operations. There exists several ways to pass a generator, and we present them next.
+
+### Tensor generator
+
+The first case of generator we can provide is simply an arbitrary hermitian tensor.
+Note we can use a string for defining the time evolution as a parameter, instead of directly passing a tensor.
+
+```python exec="on" source="material-block" html="1"
+import torch
+from pyqtorch import uniform_state, HamiltonianEvolution
+from pyqtorch.matrices import DEFAULT_MATRIX_DTYPE
+
+n_qubits = 2
+qubit_targets = list(range(n_qubits))
+
+# Random hermitian hamiltonian
+matrix = torch.rand(2**n_qubits, 2**n_qubits, dtype=DEFAULT_MATRIX_DTYPE)
+hermitian_matrix = matrix + matrix.T.conj()
+
+time = torch.tensor([1.0])
+time_symbol = "t"
+
+hamiltonian_evolution = HamiltonianEvolution(hermitian_matrix, time_symbol, qubit_targets)
+
+# Starting from a uniform state
+psi_start = uniform_state(n_qubits)
+
+# Returns the evolved state
+psi_end = hamiltonian_evolution(state = psi_start, values={time_symbol: time})
+
+print(psi_end)
+```
+
+### Symbol generator
+
+We can also have a symbol generator to be replaced later by any hermitian matrix. and in this case we use a string symbol to instantiate `HamiltonianEvolution`.
+
+```python exec="on" source="material-block" html="1"
+import torch
+from pyqtorch import uniform_state, HamiltonianEvolution
+from pyqtorch.matrices import DEFAULT_MATRIX_DTYPE
+
+n_qubits = 2
+qubit_targets = list(range(n_qubits))
+
+# Symbol hamiltonian
+hermitian_symbol = "h"
+
+time = torch.tensor([1.0])
+
+hamiltonian_evolution = HamiltonianEvolution(hermitian_symbol, time, qubit_targets)
+
+# Starting from a uniform state
+psi_start = uniform_state(n_qubits)
+
+# Set the value for h
+matrix = torch.rand(2**n_qubits, 2**n_qubits, dtype=DEFAULT_MATRIX_DTYPE)
+H = matrix + matrix.T.conj()
+
+# Returns the evolved state
+psi_end = hamiltonian_evolution(state = psi_start, values={hermitian_symbol: H})
+
+print(psi_end)
+```
+
+### Sequence generator
+
+The generator can be also a sequence of operators such as a quantum circuit:
+
+```python exec="on" source="material-block" html="1"
+import torch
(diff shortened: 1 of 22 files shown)


Roland-djee:
To be merged after #220.

jpmoutinho:
Hey @EthanObadia, closing this MR since your work is already merged to main :)

Thanks again for your contributions!

EthanObadia:
> Hey @EthanObadia, closing this MR since your work is already merged to main :)
> 
> Thanks again for your contributions!

Hi @jpmoutinho, I ended up taking a few unexpected days off, so I’ve just now had the chance to look at this branch. Thank you so much for your help with my work. 
I noticed all the reorganization that has been done, and I find that the organization of `pyq` within different files and folders is much clearer now. I also find the syntax of the two `forward` functions much clearer this way, as well as the parent class `QuantumOperation`.

PR #228, opened 2024-07-15
[Feature] Time-dependence Port

dominikandreasseitz:


pyqtorch/analog.py
```diff
@@ -522,7 +530,10 @@ def __init__(
             qubit_support: The qubits the operator acts on.
             generator_parametric: Whether the generator is parametric or not.
         """
-
+        self.is_time_dependent = is_time_dependent
+        self.solver_name = solver
```
Roland-djee, line 534:
```suggestion
        self.solver_type = solver
```

pyqtorch/analog.py
```diff
@@ -665,15 +676,50 @@ def forward(
...
             The transformed state.
         """
+        if self.is_time_dependent and embedding is not None:
+            n_qubits = len(state.shape) - 1
+            batch_size = state.shape[-1]
+            tsave = torch.linspace(0, float(self.duration), self.steps)
+
+            def Ht(t: torch.Tensor) -> torch.Tensor:
+                """Accepts a value 't' for time
+                and returns a (2**n_qubits, 2**n_qubits) Hamiltonian.
```
Roland-djee, line 686:
```suggestion
                and returns a (2**n_qubits, 2**n_qubits) Hamiltonian evaluated at 't'.
```

pyqtorch/analog.py
```diff
@@ -665,15 +676,50 @@ def forward(
         Returns:
             The transformed state.
         """
+        if self.is_time_dependent and embedding is not None:
+            n_qubits = len(state.shape) - 1
+            batch_size = state.shape[-1]
+            tsave = torch.linspace(0, float(self.duration), self.steps)
```
Roland-djee, line 682:
```suggestion
            t_grid = torch.linspace(0, float(self.duration), self.steps)
```

pyqtorch/analog.py
```diff
@@ -665,15 +676,50 @@ def forward(
...
 
-        evolved_op = self.tensor(values, None)
-        return apply_operator(
-            state=state,
-            operator=evolved_op,
-            qubits=self.qubit_support,
-            n_qubits=len(state.size()) - 1,
-            batch_size=evolved_op.shape[BATCH_DIM],
-        )
+            state = sol.states[
```
Roland-djee, line 706: Would it make sense to add a `final_state` method to `states` ? I think Pulser does it under the hood.

pyqtorch/analog.py
```diff
@@ -665,15 +676,50 @@ def forward(
...
+                -1
+            ]  # Retrieve the last state of shape (2**n_qubits, batch_size)
+
+            return state.reshape(
+                [2] * n_qubits + [batch_size]
+            )  # reshape to [2] * n_qubits + [batch_size]
+
+        else:
+            evolved_op = self.tensor(values, None)
```
Roland-djee, line 715: Is it necessary to pass `None` to `n_qubits` ? As I understood it from l. 693.

tests/test_analog.py
```diff
@@ -440,3 +440,42 @@ def test_hamevo_endianness_cnot() -> None:
     cnot_op = pyq.CNOT(0, 1)
     wf_cnot = cnot_op(state_10)
     assert torch.allclose(wf_cnot, wf_hamevo, rtol=RTOL, atol=ATOL)
+
+
+@pytest.mark.xfail  # Hello vytautas
```
Roland-djee, line 445: :)

tests/test_analog.py
```diff
@@ -440,3 +440,42 @@ def test_hamevo_endianness_cnot() -> None:
...
+        ],
+    )
+    generator = generator.tensor(
+        n_qubits=n_qubits, values={vparam: torch.tensor([0.5])}
+    )
+    generator = generator + torch.conj(torch.transpose(generator, 0, 1))
+    hamevo = pyq.HamiltonianEvolution(
+        generator, tparam, tuple(range(n_qubits)), parametric, is_time_dependent=True
+    )
+    # assert hamevo.generator_type == GeneratorType.TENSOR
```
Roland-djee, line 471: Remove.

Roland-djee (commented):
Thanks @dominikandreasseitz looking good. Some comments from me.

vytautas-a:
@kaosmicadei, here's the working version of time-dependent generator for `HamiltonianEvolution`.

pyqtorch/hamiltonians/evolution.py
```diff
@@ -331,13 +338,50 @@ def forward(
...
-            state=state,
-            operator=evolved_op,
-            qubit_support=self.qubit_support,
-        )
+            # Retrieve the last state of shape (2**n_qubits, batch_size)
+            state = sol.states[-1]
+
+            return state.reshape(
+                [2] * n_qubits + [batch_size]
+            )  # reshape to [2] * n_qubits + [batch_size]
```
jpmoutinho, line 378: I don't think we need to keep all the intermediate comments

pyqtorch/utils.py
```diff
@@ -741,3 +743,15 @@ class SolverType(StrEnum):
...
+def is_parametric(operation: pyq.Sequence) -> bool:
+    params = []
+    for m in operation.modules():
+        if isinstance(m, pyq.Scale):
+            params.append(m.param_name)
+
+    res = False
+    if any([isinstance(p, str) for p in params]):
+        res = True
+    return res
```
jpmoutinho, line 757: Nice :) I think it could be useful to extend this function to work for any operation in PyQ.

jpmoutinho (commented):
Very nice @vytautas-a, look good :)

I would just do some code cleanup on the `HamEvo` class and then it should be good to go.

pyqtorch/utils.py
vytautas-a: I think it should work already now for any operation - the only place where a non-fixed parameter may sit is the `Scale` block. So that's what is checked here.

jpmoutinho: Hmm I think this would fail for something like `Sequence([RX, RY])` or even just a single RX. Ok to leave it like that for now, but in which case I would add some info in the docstring.

pyqtorch/hamiltonians/evolution.py
vytautas-a: done

jpmoutinho (approved):
Thanks @vytautas-a. Feel free to merge in the meantime and then we can release. I've already bumped to 1.4.7 here: https://github.com/pasqal-io/pyqtorch/pull/276

pyqtorch/utils.py
```diff
@@ -741,3 +743,15 @@ class SolverType(StrEnum):
 
     KRYLOV_SE = "krylov_se"
     """Uses Krylov Schrodinger equation solver"""
+
+
+def is_parametric(operation: pyq.Sequence) -> bool:
```
Roland-djee, line 748: I would import these types from PyQ, just for the sake of readability.

pyqtorch/utils.py
```diff
@@ -741,3 +743,15 @@ class SolverType(StrEnum):
...
+    params = []
+    for m in operation.modules():
+        if isinstance(m, pyq.Scale):
+            params.append(m.param_name)
+
+    res = False
+    if any([isinstance(p, str) for p in params]):
```
Roland-djee, line 757:
```suggestion
    if any(isinstance(p, str) for p in params):
```

pyqtorch/utils.py
vytautas-a: Yeah, that won't work now, but the solution is easy - just check in `isinstance` for `Parametric` class as well. That should cover all the parametrized operations.

pyqtorch/hamiltonians/evolution.py
```diff
@@ -315,6 +322,55 @@ def spectral_gap(self) -> Tensor:
...
+        state: Tensor,
+        values: dict[str, Tensor] | ParameterDict = dict(),
+        embedding: Embedding | None = None,
+    ) -> State:
+        evolved_op = self.tensor(values, embedding)
+        return apply_operator(
+            state=state, operator=evolved_op, qubit_support=self.qubit_support
+        )
+
+    def _forward_time(
```
Roland-djee, line 336: I assume it wasn't straightforward to merge it with the `_forward` above ?

Roland-djee (approved):
Thanks @vytautas-a. My own two cents from afar.

pyqtorch/utils.py
vytautas-a: yes, I tried that but the circular import error didn't allow that. I simply didn't want to fight that here

pyqtorch/hamiltonians/evolution.py
vytautas-a: well, this refactoring is the direct consequence of @jpmoutinho comment from above that suggested separating both cases (time-dependent and independent) into different functions. Before this refactoring everything was in `forward()`

Roland-djee: OK no worries. I was just curious about the design.

pyqtorch/utils.py
Roland-djee: hm OK. Thanks.

PR #230, opened 2024-07-16
[Feature] MeasurementProtocols class for finite shot handling

chMoussa:
Just framing the MeasurementProtocols class for shot handling when computing expectation values.


chMoussa:
Eventually, this will not be added in pyqtorch.

PR #233, opened 2024-07-16
[Bug] Fix check support PSR

chMoussa:
Fix the `if` order in `check_support_psr` to avoid checking the spectral gap for other unconcerned gates, which raises an error.


dominikandreasseitz (approved):
thanks!

PR #234, opened 2024-07-16
[Testing] Increase cross-platform embedding atol

chMoussa:
The pipeline may fail at precision `1e-06`. So reducing it to 3.


dominikandreasseitz (approved):
quick and easy! thanks @chMoussa

PR #239, opened 2024-07-22
[Feature] Add finite shots expectation value

chMoussa:
Shots handling include:
- [x] Defining an expectation method to compute expectations via sampling using the eigendecomposition of the observable
- [x] Make it possible to use the sampled expectation for `AD` and `GPSR` forward mode.
- [x] Use the method also in the `PSRExpectation` backward by using the method as argument of forward, and saving in `ctx`. 

Note for later: To handle dtype complex64, we need to set up properly the dtype in the tensor method of a few operators. To be synced with #220


pyqtorch/api.py
```diff
@@ -96,12 +99,98 @@ def sample(
...
 
 
+def analytical_expectation(
+    circuit: QuantumCircuit,
+    state: Tensor,
+    observable: Observable,
+    values: dict[str, Tensor] = dict(),
+    embedding: Embedding | None = None,
+) -> Tensor:
+    """Non sampled expectation value.
```
Roland-djee, line 109:
```suggestion
    """Compute the analytical expectation value.
```

pyqtorch/api.py
```diff
@@ -96,12 +99,98 @@ def sample(
...
+    )
+    eigvals = eigvals.squeeze()
+    eigvec_state_prod = apply_operator(
+        state,
+        eigvecs.T.conj(),
+        tuple(range(n_qubits)),
+        n_qubits=circuit.n_qubits,
+    )
+    eigvec_state_prod = torch.flatten(eigvec_state_prod, start_dim=0, end_dim=-2).t()
+    probs = torch.abs(torch.pow(eigvec_state_prod, 2))
```
Roland-djee, line 169: Shouldn't that be the absolute squared instead ?

pyqtorch/api.py
```diff
@@ -96,12 +99,98 @@ def sample(
...
+    return expectations
+
+
 def expectation(
     circuit: QuantumCircuit,
     state: Tensor = None,
     values: dict[str, Tensor] = dict(),
     observable: Observable = None,  # type: ignore[assignment]
     diff_mode: DiffMode = DiffMode.AD,
+    options: dict[str, int] = dict(),
```
Roland-djee, line 193: Could it contain any other option than the number of shots ?

pyqtorch/api.py
```diff
@@ -144,13 +237,20 @@ def expectation(
     )
     if observable is None:
         logger.error("Please provide an observable to compute expectation.")
+
     if state is None:
         state = circuit.init_state(batch_size=1)
+
+    expectation_fn = analytical_expectation
+    n_shots = options.get("n_shots")
```
Roland-djee, line 245:
```suggestion
    n_shots = options.get("n_shots", None)
```

pyqtorch/utils.py
```diff
@@ -69,6 +71,44 @@ def overlap(bra: Tensor, ket: Tensor) -> Tensor:
     return torch.pow(inner_prod(bra, ket).real, 2)
 
 
+def sample_multinomial(
+    probs: Tensor,
+    length_bitstring: int,
+    n_samples: int,
+    return_counter: bool = True,
+    minlength: int = 0,
+) -> Union[Counter, Tensor]:
```
Roland-djee, line 80:
```suggestion
) -> Counter | Tensor:
```

pyqtorch/utils.py
```diff
@@ -69,6 +71,44 @@ def overlap(bra: Tensor, ket: Tensor) -> Tensor:
     return torch.pow(inner_prod(bra, ket).real, 2)
 
 
+def sample_multinomial(
+    probs: Tensor,
+    length_bitstring: int,
+    n_samples: int,
+    return_counter: bool = True,
```
Roland-djee, line 78: Do we want to let the user decide on the return type ? It feels simpler and easier to maintain to agree on a single one.

tests/test_differentiation.py
```diff
@@ -65,6 +65,54 @@ def test_adjoint_diff(n_qubits: int, n_layers: int) -> None:
     # )
 
 
+@pytest.mark.parametrize("n_qubits", [3])
+@pytest.mark.parametrize("batch_size", [1])
```
Roland-djee, line 69: Is it worth parametrising for a single value ?

tests/test_differentiation.py
```diff
@@ -65,6 +65,54 @@ def test_adjoint_diff(n_qubits: int, n_layers: int) -> None:
...
+@pytest.mark.parametrize("ops_op", [pyq.Z, pyq.Y])
+@pytest.mark.parametrize("dtype", [torch.complex128])
```
Roland-djee, line 71: Ditto.

Roland-djee (commented):
Thanks a lot @chMoussa. Some minor comments from me. Will approve once addressed.

pyqtorch/api.py
chMoussa: @Roland-djee At the moment any extra information will be unused but I left it as it is to be flexible later, say we can pass a post-processing method or a custom expectation method. Would you suggest we just use for now `n_shots` as an argument of `expectation` instead of the `options` dict, and later adapt for future options?

pyqtorch/utils.py
chMoussa: @Roland-djee I made it this way to factorize this function as `circuit.sample` would return the counter, and for the sampled expectation case with eigendecomposition, the tensor is needed for computation.

tests/test_differentiation.py
chMoussa: Oups, forgot to put more values back...

tests/test_differentiation.py
chMoussa: I will add complex64 whenever we are sync with #220

pyqtorch/api.py
chMoussa: Indeed. I got tricked by the same mistake in circuit.sample.

tests/test_differentiation.py
```diff
@@ -65,6 +65,54 @@ def test_adjoint_diff(n_qubits: int, n_layers: int) -> None:
...
 
 
+@pytest.mark.parametrize("n_qubits", [3, 5])
+@pytest.mark.parametrize("batch_size", [1, 2])
+@pytest.mark.parametrize("ops_op", [pyq.Z, pyq.Y])
+@pytest.mark.parametrize("dtype", [torch.complex128])
+def test_sampled_diff(
+    n_qubits: int,
+    batch_size: int,
+    ops_op: Primitive,
```
Roland-djee, line 75:
```suggestion
    ops_op: type[Primitive],
```

Tbc though.

tests/test_gpsr.py
```diff
@@ -78,7 +78,7 @@ def circuit_sequence(n_qubits: int) -> QuantumCircuit:
     ],
 )
 @pytest.mark.parametrize("ops_op", [pyq.Z, pyq.X, pyq.Y])
-@pytest.mark.parametrize("dtype", [torch.complex64, torch.complex128])
+@pytest.mark.parametrize("dtype", [torch.complex128])
```
Roland-djee, line 81: Not sure it is useful for a single parameter.

Roland-djee (commented):
Thanks for the updates @chMoussa. Final comments from my side and I think we can ship it over once addressed.

pyqtorch/api.py
```diff
@@ -96,12 +99,98 @@ def sample(
...
+    return expectations
+
+
 def expectation(
     circuit: QuantumCircuit,
     state: Tensor = None,
     values: dict[str, Tensor] = dict(),
     observable: Observable = None,  # type: ignore[assignment]
     diff_mode: DiffMode = DiffMode.AD,
+    n_shots: int | None = None,
```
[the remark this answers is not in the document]
chMoussa: No, cause it is what decide whether we run expectation using sampling or via matmul.

pyqtorch/api.py
```diff
@@ -144,13 +236,19 @@ def expectation(
     )
     if observable is None:
         logger.error("Please provide an observable to compute expectation.")
+
     if state is None:
         state = circuit.init_state(batch_size=1)
+
+    expectation_fn = analytical_expectation
+    if n_shots is not None:
```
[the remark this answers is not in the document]
chMoussa: Same as above comment.

tests/test_gpsr.py
chMoussa: torch.complex64 will be added later, after the tensor modifications brought by #220 are in

tests/test_differentiation.py
chMoussa: pre-commit does not like it, keeping Primitive for now

tests/test_gpsr.py
chMoussa: It gives an error due to missing the dtype argument when creating identities, but we will fix directly after the MR.

pyqtorch/api.py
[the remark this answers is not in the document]
chMoussa: In other places, I set it to 1 due to knowing beforehand this will be used with shots.

pyqtorch/api.py
```diff
@@ -96,12 +99,98 @@ def sample(
...
+    eigvals, eigvecs = torch.linalg.eigh(
+        observable.tensor(values=values, embedding=embedding, full_support=list(range(n_qubits))).permute((2, 0, 1))
+    )
+    eigvals = eigvals.squeeze()
+    eigvec_state_prod = apply_operator(
+        state,
+        eigvecs.T.conj(),
+        tuple(range(n_qubits)),
+        n_qubits=circuit.n_qubits,
+    )
```
jpmoutinho, line 167: You can make this more efficient by doing `observable.tensor()` without the `full_support`, calling `eigh`, and then expanding the `eigenvalues` and `eigenvectors` if needed. See the comment I made [here](https://github.com/pasqal-io/horqrux/pull/23#discussion_r1692698274) on horqrux (actually just today xD).

jpmoutinho: After you get the `eigvecs` matrix it might be enough to call `expand_operator` on it, but best to double-check this.

tests/test_differentiation.py
```diff
@@ -65,6 +65,52 @@ def test_adjoint_diff(n_qubits: int, n_layers: int) -> None:
...
+    dtype: torch.dtype,
+) -> None:
+    rx = pyq.RX(0, param_name="theta_0")
+    cry = pyq.CPHASE(0, 1, param_name="theta_1")
+    rz = pyq.RZ(2, param_name="theta_2")
+    cnot = pyq.CNOT(1, 2)
+    ops = [rx, cry, rz, cnot]
+    circ = pyq.QuantumCircuit(n_qubits, ops).to(dtype)
+    obs = pyq.Observable(pyq.Add([ops_op(i) for i in range(n_qubits)])).to(dtype)
```
jpmoutinho, line 84: You can also try to use the `random_pauli_hamiltonian` in `helpers.py`.

chMoussa: Using `random_pauli_hamiltonian`, the tests fail with hamiltonians composed of both 1q and 2q terms. Investigating this

pyqtorch/api.py
chMoussa: @jpmoutinho I think I will tackle this in another MR. This requires a subtle logic for the eigenvalues as well.

tests/test_differentiation.py
chMoussa: This is due to the error scaling with the observable definition as well as the number of shots. Changing the test now for testing simpler cases of using coeffs 1.0 with Scale in gpsr test.

pyqtorch/primitive.py
```diff
@@ -302,6 +302,7 @@ def __init__(self, control: int | tuple[int, ...], target: int):
 
 
 OPS_PAULI = {X, Y, Z, I, N}
+OPS_PAULI_with_generator = {X, Y, Z, I}
```
jpmoutinho, line 305: Just remove `N` from ops `OPS_PAULI`, it's more correct that way anyway, and it doesn't really affect much the tests since it is just a combination of `I` and `Z`.

jpmoutinho (approved):

Issue #241, opened 2024-07-26
[Refact] Optimize `permute_basis` function


jpmoutinho:
The goal of the `permute_basis` function is to permute the rows and columns of a given operator matrix according to a permutation on the qubit support that is different than `tuple(range(n_qubits))`. E.g., the matrix for `CNOT(0, 1)` is
```
tensor([[1., 0., 0., 0.],
        [0., 1., 0., 0.],
        [0., 0., 0., 1.],
        [0., 0., 1., 0.]], dtype=torch.float64)
```
but the one for `CNOT(1, 0)` should be
```
tensor([[1., 0., 0., 0.],
        [0., 0., 0., 1.],
        [0., 0., 1., 0.],
        [0., 1., 0., 0.]], dtype=torch.float64)
```
Currently this permutation is generated by explicitly writing down all the basis states in bitstring format, permuting the position of all qubits according to the qubit support order, and then rebuilding the integer values using the different order of the basis states, which gives the correct basis state permutation that can be used to slice the operator with. This is handled by the `_basis_order` function in `pyqtorch.bitstrings`.

As @chMoussa pointed out [here](https://github.com/pasqal-io/pyqtorch/pull/220#discussion_r1689394629), there is probably a smarter way of doing it with `torch.permute`. This requires viewing the operator with a similar shape to what is done in the function `apply_operator`, then permuting the dimensions of the tensor according to the qubit support, and then reshaping back. Below is a trial code to do that:

```
def new_permute(operator: torch.Tensor, qubit_support: tuple[int, ...]) -> torch.Tensor:
    n_qubits = len(qubit_support)
    batch_size = operator.size(-1)
    operator = operator.view([2] * 2 * n_qubits + [batch_size])
    perm = list(qubit_support + tuple(n_qubits + q for q in qubit_support) + (2*n_qubits,))
    return operator.permute(perm).reshape([2**n_qubits, 2**n_qubits, batch_size])
```

This *almost* does it, but it is still not fully working. For example, see the code below:

```
import torch
import random
from pyqtorch.bitstrings import permute_basis

n_qubits = 3
operator = torch.randint(2, (2**n_qubits, 2**n_qubits)).unsqueeze(2)

def new_permute(operator, qubit_support):
    n_qubits = len(qubit_support)
    batch_size = operator.size(-1)
    operator = operator.view([2] * 2 * n_qubits + [batch_size])
[...]

PR #242, opened 2024-07-29
[Refact] Optimize permute_basis function

chMoussa:
Solves #241


pyqtorch/utils.py
```diff
@@ -414,6 +413,27 @@ def promote_operator(operator: Tensor, target: int, n_qubits: int) -> Tensor:
...
+        Tensor: Permuted operator.
+    """
+    n_qubits = len(qubit_support)
+    batch_size = operator.size(-1)
+    operator = operator.view([2] * 2 * n_qubits + [batch_size])
+    ordered_support = argsort(qubit_support)
+    perm = list(
+        tuple(ordered_support) + tuple(ordered_support + n_qubits) + (2 * n_qubits,)
+    )
+    return operator.permute(perm).reshape([2**n_qubits, 2**n_qubits, batch_size])
```
jpmoutinho, line 434: Can you add a check that if `qubit_support == tuple(sorted(qubit_support))` it does early return of the operator? That way we for sure avoid calling the `permute` and `reshape` if not needed.

jpmoutinho (changes requested):

pyqtorch/utils.py
jpmoutinho: Possibly those functions are also smart not to do anything if they don't need to.

jpmoutinho (approved):

PR #255, opened 2024-08-07
[Feature] HamiltonianEvolution supported in GPSR with fixed generators

chMoussa:


pyqtorch/matrices.py
```diff
@@ -67,14 +67,38 @@ def PROJMAT(ket: Tensor, bra: Tensor) -> Tensor:
...
 def parametric_unitary(
-    theta: torch.Tensor, P: torch.Tensor, I: torch.Tensor, batch_size: int  # noqa: E741
+    theta: torch.Tensor,
+    P: torch.Tensor,
+    identity_mat: torch.Tensor,
+    batch_size: int,
+    a: float = 0.5,  # noqa: E741
 ) -> torch.Tensor:
-    cos_t = torch.cos(theta / 2).unsqueeze(0).unsqueeze(1)
+    """Compute the exponentiation of a matrix :math:`P`
```
jpmoutinho, line 76: exponentiation of a _Pauli_ matrix right?

pyqtorch/matrices.py
```diff
@@ -67,14 +67,38 @@ def PROJMAT(ket: Tensor, bra: Tensor) -> Tensor:
 
 
 def parametric_unitary(
-    theta: torch.Tensor, P: torch.Tensor, I: torch.Tensor, batch_size: int  # noqa: E741
+    theta: torch.Tensor,
+    P: torch.Tensor,
+    identity_mat: torch.Tensor,
```
jpmoutinho, line 72: Why do you need this as an argument? Can't you use `IMAT` for `pyqtorch.matrices`?

pyqtorch/primitive.py
jpmoutinho, line 1: Seems you have this duplicated from the new `pyqtorch.primitives` module.

Roland-djee (approved):
LGTM. Not much to add from @jpmoutinho's comments.

pyqtorch/matrices.py
chMoussa: It is for supporting the device and dtype conversions from the Parametric gate.

PR #260, opened 2024-08-12
[Bugfix] Divide shift by spectral gap for single-gap PSR

chMoussa:
To account for the factor 2 between the definition of `HamiltonianEvolution` and usual `Parametric` gates, one needs to rescale the shift with the spectral gap on top of dividing by 2.


jpmoutinho (approved):
Thanks @chMoussa :) Once you coverge on all these details please write some notes on it. We can put it in the docs.

PR #262, opened 2024-08-13
[Bug, Feature] Fix permute basis and add permutation based apply operator

jpmoutinho:
A few things that came up while working on the improvement to the noise.

- The `permute_basis` was not doing the correct permutation, but it was going unnoticed because the `expand_operator` was compensating for it.
- I created an alternative `apply_operator_permute` which is just as fast as the `apply_operator` function, but can serve as a basis for future changes. For now we can simply save it. It could be useful if we wish to make PyQ follow the more standard state shape of `[batch_size, 2**n_qubits]` instead of the `[2] * n_qubits + [batch_size]`. The reason I wrote it is because the logic I am working on for the `apply_density_mat` follows something similar, but doing it first for the normal `apply_operator` was easier.

### DIFF
--- a/pyqtorch/apply.py
+++ b/pyqtorch/apply.py
@@ -7,7 +7,7 @@ from numpy.typing import NDArray
 from torch import Tensor, einsum
 
 from pyqtorch.matrices import _dagger
-from pyqtorch.utils import DensityMatrix
+from pyqtorch.utils import DensityMatrix, permute_state
 
 ABC_ARRAY: NDArray = array(list(ABC))
 
@@ -15,9 +15,7 @@ ABC_ARRAY: NDArray = array(list(ABC))
 def apply_operator(
     state: Tensor,
     operator: Tensor,
-    qubits: tuple[int, ...] | list[int],
-    n_qubits: int | None = None,
-    batch_size: int | None = None,
+    qubit_support: tuple[int, ...] | list[int],
 ) -> Tensor:
     """Applies an operator, i.e. a single tensor of shape [2, 2, ...], on a given state
        of shape [2 for _ in range(n_qubits)] for a given set of (target and control) qubits.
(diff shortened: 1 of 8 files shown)


pyqtorch/apply.py
```diff
@@ -7,17 +7,15 @@
...
 
 
 def apply_operator(
     state: Tensor,
     operator: Tensor,
-    qubits: tuple[int, ...] | list[int],
-    n_qubits: int | None = None,
-    batch_size: int | None = None,
+    qubit_support: tuple[int, ...] | list[int],
 ) -> Tensor:
```
jpmoutinho, line 19: I also simplified this interface since the extra information was not needed.

tests/test_tensor.py
```diff
@@ -49,10 +49,13 @@ def test_digital_tensor(n_qubits: int, batch_size: int, use_full_support: bool)
...
 
 
+@pytest.mark.parametrize("use_permute", [True, False])
 @pytest.mark.parametrize("use_full_support", [True, False])
 @pytest.mark.parametrize("n_qubits", [4, 5])
 @pytest.mark.parametrize("batch_size", [1, 5])
-def test_param_tensor(n_qubits: int, batch_size: int, use_full_support: bool) -> None:
+def test_param_tensor(
+    n_qubits: int, batch_size: int, use_full_support: bool, use_permute: bool
+) -> None:
```
jpmoutinho, line 58: I did test running all the `tensor` tests with the `apply_operator_permute` and they all pass, but I figured it would be too much spam to run them all everytime in every pipeline.

pyqtorch/utils.py
```diff
@@ -445,7 +446,34 @@ def promote_operator(operator: Tensor, target: int, n_qubits: int) -> Tensor:
...
+def permute_state(
+    state: Tensor, qubit_support: tuple | list, inv: bool = False
+) -> Tensor:
+    """Takes a state tensor and permutes the qubit amplitudes
+    according to the order of the qubit support.
+
+    Args:
+        state (Tensor): State to permute over.
+        qubit_support (tuple): Qubit support.
+
```
chMoussa, line 458: Missing inv description in docstring

chMoussa (approved):

PR #264, opened 2024-08-15
[Bug] Fix control ops tensor and qubit permutations

jpmoutinho:


chMoussa (approved):
Thanks @jpmoutinho! lgtm

PR #265, opened 2024-08-16
[Feature] Add jacobian method for adjoint supporting hamevo

chMoussa:


pyqtorch/hamiltonians/evolution.py
```diff
@@ -372,3 +373,27 @@ def tensor(
...
+
+        Returns:
+            The unitary representation of the jacobian.
+        """
+
+        hamiltonian: torch.Tensor = self.create_hamiltonian(values, embedding)  # type: ignore [call-arg]
+        time_evolution: torch.Tensor = (
+            values[self.time] if isinstance(self.time, str) else self.time
+        )  # If `self.time` is a string / hence, a Parameter,
+        # we expect the user to pass it in the `values` dict
```
Doomsk, line 394: Maybe it's worth checking the `self.time` type (although it's annotated in the class `__init__` method that it can be `str` or `Tensor`, nothing prevents it not being. So either enforcing that on instantiation or here would be good to spot early errors.

pyqtorch/utils.py
```diff
@@ -221,6 +222,57 @@ def is_diag(H: Tensor, atol: Tensor = ATOL) -> bool:
...
+        order = len(derivative_indices)
+        eps = torch.finfo(x.dtype).eps ** (1 / (2 + order))
+
+    # compute derivative direction vector(s)
+    eps = torch.as_tensor(eps, dtype=x.dtype)
+    _eps = 1 / eps  # type: ignore[operator]
+    ev = torch.zeros_like(x)
+    i = derivative_indices[0]
+    ev[:, i] += eps
+
```
Doomsk, line 255: Since `eps` seems to not be used anymore in the code, maybe having `ev[:, i] := eps` first and then giving a meaningful variable name for `1 / eps` operation (what is even `ev`?) to favor code description

pyqtorch/utils.py
```diff
@@ -221,6 +222,57 @@ def is_diag(H: Tensor, atol: Tensor = ATOL) -> bool:
...
+        )
+    elif len(derivative_indices) == 2:
+        return (f(x + ev) + f(x - ev) - 2 * f(x)) * _eps**2
+    elif len(derivative_indices) == 1:
+        return (f(x + ev) - f(x - ev)) * _eps / 2
+    else:
+        raise ValueError(
+            "If you see this error there is a bug in the `finitediff` function."
+        )
+
```
Doomsk, line 274: if your `if` statement has a `return` inside, you may consider using the `elif` statements as `if`'s (because they'll be naturally called, should the previous `if` return false, e.g.:
```suggestion
    if len(derivative_indices) > 3 or len(set(derivative_indices)) > 1:
        di = derivative_indices[1:]
        return (finitediff(f, x + ev, di) - finitediff(f, x - ev, di)) * _eps / 2
    if len(derivative_indices) == 3:
        return (
            (f(x + 2 * ev) - 2 * f(x + ev) + 2 * f(x - ev) - f(x - 2 * ev))
            * _eps**3
            / 2
        )
    if len(derivative_indices) == 2:
        return (f(x + ev) + f(x - ev) - 2 * f(x)) * _eps**2
    if len(derivative_indices) == 1:
        return (f(x + ev) - f(x - ev)) * _eps / 2
    raise ValueError(
            "If you see this error there is a bug in the `finitediff` function."
        )

```

tests/test_differentiation.py
Doomsk, line 1: I failed to find the use of the `jacobian` method here; where should I look for it?

Doomsk (changes requested):
the code seems good! A few comments the may be helpful and others that are relevant to see how these changes affects the overall code

tests/test_differentiation.py
chMoussa: It is used in the adjoint method to calculate gradients wrt circuit parameters.

pyqtorch/utils.py
chMoussa: Good remarks, I took it from qadence, ev represents the step/delta for differentiation. Will rename accordingly.

Doomsk (approved):
Great! I finally got the example (it was lacking for hamevos, but the rest was already there)

PR #268, opened 2024-08-22
[Performance, Refactor] Operator multiplication without explicit identities

jpmoutinho:
Refactors the `operator_product` function that does not require explicit identity padding. Also refactors the `apply_density_mat` into a new `apply_operator_dm` without explicit identities. The logic of both is similar, and the logic of `apply_operator_dm` is very similar to `apply_operator_permute`, but keeping everything separate for now.

The reason not to use `operator_product` inside `apply_operator_dm` is to avoid permuting the qubits twice.

The new functions are similar in performance for small number of qubits. For 5 to 9 qubits they become a few times faster. For 12 qubits the speedup is already ~10x.


jpmoutinho:
Hey @EthanObadia finally got around to writing the `apply_operator` equivalent for density matrices without using explicit identity padding. If you have some time to check it out it would be great :)

chMoussa (approved):

Issue #267, opened 2024-08-22
[Feature] Add issue template


chMoussa:

chMoussa:
Closes with #269

PR #269, opened 2024-08-23
[Feature] Add issue template

chMoussa:
Solves #267


jpmoutinho (approved):
Thanks :) 

Btw can you bump the patch? Will need a new release to implement the noise in Qadence.

PR #275, opened 2024-08-30
[Feature, Testing] Allow nested calls to ConcretizedCallable, more tests

jpmoutinho:
Extends the `ConcretizedCallable` to allow nested calls, adds some convenience functions. E.g. now we can write

```
import pyqtorch as pyq

## Creating an expression and embedding

expr = "z" ** pyq.log(1.0 / (1.0 + (2.0 * pyq.sin("x")) + "y"))

expr_name = "expr"

embedding = pyq.Embedding(
    fparam_names = ["x", "y"],
    vparam_names = ["z"],
    var_to_call = {"expr": expr}
)

## Passing the expression to a block parameter

op = pyq.Scale(pyq.I(0), "expr")

values = {"x": torch.tensor(1.0), "y": torch.tensor(-1.0),  "z": torch.tensor(2.0)}

matrix = op.tensor(values, embedding)

matrix[..., 0]

---

tensor([[0.6971, 0.0000],
        [0.0000, 0.6971]])
```

Furthermore, with the current change, we might not need the `Embedding` 🤔. To be explored in a later MR: https://github.com/pasqal-io/pyqtorch/issues/278

### DIFF
--- a/pyqtorch/composite/compose.py
+++ b/pyqtorch/composite/compose.py
@@ -10,7 +10,7 @@ from torch import Tensor, einsum, rand
 from torch.nn import Module, ModuleList, ParameterDict
 
 from pyqtorch.apply import apply_operator
-from pyqtorch.embed import ConcretizedCallable, Embedding
+from pyqtorch.embed import Embedding
 from pyqtorch.matrices import add_batch_dim
 from pyqtorch.primitives import CNOT, RX, RY, Parametric, Primitive
 from pyqtorch.utils import (
@@ -37,7 +37,7 @@ class Scale(Sequence):
     def __init__(
         self,
         operations: Union[Primitive, Sequence, Add],
-        param_name: str | Tensor | ConcretizedCallable,
+        param_name: str | Tensor,
     ):
         """
         Initializes a Scale object.
@@ -75,8 +75,6 @@ class Scale(Sequence):
             scale = values[self.param_name]
         elif isinstance(self.param_name, Tensor):
             scale = self.param_name
-        elif isinstance(self.param_name, ConcretizedCallable):
-            scale = self.param_name(values)
 
         return scale * self.operations[0].forward(state, values)
 
@@ -105,8 +103,6 @@ class Scale(Sequence):
             scale = values[self.param_name]
         elif isinstance(self.param_name, Tensor):
             scale = self.param_name
-        elif isinstance(self.param_name, ConcretizedCallable):
-            scale = self.param_name(values)
 
         return scale * self.operations[0].tensor(values, full_support=full_support)
 
--- a/pyqtorch/composite/compose.py
+++ b/pyqtorch/composite/compose.py
@@ -76,8 +76,7 @@ class Scale(Sequence):
             if isinstance(self.param_name, str)
             else self.param_name
         )
-
-        return scale * self.operations[0].forward(state, values)
+        return scale * self.operations[0].forward(state, values, embedding)
 
     def tensor(
         self,
@@ -105,7 +104,7 @@ class Scale(Sequence):
             if isinstance(self.param_name, str)
             else self.param_name
         )
-        return scale * self.operations[0].tensor(values, embedding, full_support=full_support)
+        return scale * self.operations[0].tensor(values, embedding, full_support)
 
     def flatten(self) -> list[Scale]:
         """This method should only be called in the AdjointExpectation,
--- a/pyqtorch/composite/compose.py
+++ b/pyqtorch/composite/compose.py
@@ -71,10 +71,11 @@ class Scale(Sequence):
         if embedding is not None:
             values = embedding(values)
 
-        if isinstance(self.param_name, str):
-            scale = values[self.param_name]
-        elif isinstance(self.param_name, Tensor):
-            scale = self.param_name
+        scale = (
+            values[self.param_name]
+            if isinstance(self.param_name, str)
+            else self.param_name
+        )
 
         return scale * self.operations[0].forward(state, values)
 
@@ -99,10 +100,11 @@ class Scale(Sequence):
         if embedding is not None:
             values = embedding(values)
 
-        if isinstance(self.param_name, str):
-            scale = values[self.param_name]
-        elif isinstance(self.param_name, Tensor):
-            scale = self.param_name
+        scale = (
+            values[self.param_name]
+            if isinstance(self.param_name, str)
+            else self.param_name
+        )
 
         return scale * self.operations[0].tensor(values, full_support=full_support)
 
--- a/pyqtorch/composite/compose.py
+++ b/pyqtorch/composite/compose.py
@@ -105,8 +105,7 @@ class Scale(Sequence):
             if isinstance(self.param_name, str)
             else self.param_name
         )
-
-        return scale * self.operations[0].tensor(values, full_support=full_support)
+        return scale * self.operations[0].tensor(values, embedding, full_support=full_support)
 
     def flatten(self) -> list[Scale]:
         """This method should only be called in the AdjointExpectation,
--- a/pyqtorch/composite/compose.py
+++ b/pyqtorch/composite/compose.py
@@ -10,7 +10,7 @@ from torch import Tensor, einsum, rand
 from torch.nn import Module, ModuleList, ParameterDict
 
 from pyqtorch.apply import apply_operator
-from pyqtorch.embed import Embedding
+from pyqtorch.embed import ConcretizedCallable, Embedding
 from pyqtorch.matrices import add_batch_dim
 from pyqtorch.primitives import CNOT, RX, RY, Parametric, Primitive
 from pyqtorch.utils import (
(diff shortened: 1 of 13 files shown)


pyqtorch/utils.py
```diff
@@ -218,9 +218,7 @@ def is_diag(H: Tensor, atol: Tensor = ATOL) -> bool:
     Returns:
         True if diagonal, else False.
     """
-    m = H.shape[0]
-    p, q = H.stride()
-    offdiag_view = torch.as_strided(H[:, 1:], (m - 1, m), (p + q, q))
```
jpmoutinho, line 223: Why did you need this change?

The `is_diag` function had been refactored at some point by @chMoussa and I think the `stride` thing was there because it was especially efficient.

jpmoutinho (commented):
Thanks @vytautas-a :)

I think my main question is regarding the logic on how to write `HamiltonianEvolution`, and which are the correct combinations of parameters to write a time-dependent one, time-independent one, etc.

pyqtorch/utils.py
vytautas-a: I did it actually because I got a weird runtime error with one test in `qadence`. So I refactored this into such simpler form so that it passes.

pyqtorch/hamiltonians/evolution.py
```diff
@@ -142,9 +143,10 @@ class HamiltonianEvolution(Sequence):
     def __init__(
         self,
         generator: TGenerator,
-        time: Tensor | str,
+        time: Tensor | str | ConcretizedCallable,
         qubit_support: Tuple[int, ...] | None = None,
         cache_length: int = 1,
+        duration: float | Tensor = 1.0,
         steps: int = 100,
         solver=SolverType.DP5_SE,
```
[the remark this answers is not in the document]
vytautas-a: I remember that I got away without `duration` at some point but then again reverted this decision. I can take a look once more at this.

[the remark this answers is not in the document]
vytautas-a: I looked into this issue and it seems that `duration` is unfortunately necessary because `time` has to denote the actual time parameter, thus there has to be an argument to pass the actual duration.

pyqtorch/utils.py
chMoussa: What was the weird error @vytautas-a ?

vytautas-a: Something about not being able to fit a tensor into designated memory.

vytautas-a: `FAILED tests/qadence/test_quantum_model.py::test_quantum_model_with_hevo - RuntimeError: setStorage: sizes [15, 16], strides [17, 16], storage offset 2320, and itemsize 8 requiring a storage size of 22392 are out of bounds for storage of size 20480`

pyqtorch/hamiltonians/evolution.py
```diff
@@ -246,6 +257,22 @@ def flatten(self) -> ModuleList:
...
+            return res
+        else:
+            if isinstance(generator, Sequence):
+                for m in generator.modules():
+                    if isinstance(m, (Scale, Parametric)):
+                        if self.time in getattr(m.param_name, "independent_args", []):
+                            res = True
+                        elif m.param_name == self.time:
+                            res = True
+        return res
```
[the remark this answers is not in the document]
jpmoutinho: Ah ok I think I got it. I think this deserves a test that runs each of the possible combinations of inputs and checks the code handles it correctly.

pyqtorch/primitives/parametric.py
```diff
@@ -34,7 +34,7 @@ def __init__(
         self,
         generator: str | Tensor,
         qubit_support: int | tuple[int, ...] | Support,
-        param_name: str | int | float | torch.Tensor = "",
+        param_name: str | int | float | torch.Tensor | ConcretizedCallable = "",
```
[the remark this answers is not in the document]
jpmoutinho: What I meant is that you added the support for passing a `ConcretizedCallable` directly as the parameter, as you write there in the type hint. You also added it above in the `Scale`. When I tried it, the scale worked fine, but the `Parametric` had some error that I don't recall exactly what it was. But since you added it for both, there should be a test for it.

pyqtorch/utils.py
```diff
@@ -752,8 +750,13 @@ def is_parametric(operation: pyq.Sequence) -> bool:
...
-    if any(isinstance(p, str) for p in params):
+    if any(isinstance(p, (str, pyq.ConcretizedCallable)) for p in params):
         res = True
     return res
+
+
+def heaviside(x: Tensor, _: Tensor) -> Tensor:
+    a = torch.zeros(2)
+    a[0] = x
+    return torch.clamp(1000 * torch.max(a), torch.tensor(0.0), torch.tensor(1.0))
```
jpmoutinho, line 762: Can you add a docstring explaining why this is needed, and add some tests that cover the intended use?

vytautas-a: Removed the `ConcretizedCallable` stuff from `parametric.py`. It's not actually needed

pyqtorch/primitives/parametric.py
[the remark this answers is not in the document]
vytautas-a: Removed the `ConcretizedCallable` from `parametric.py`

pyqtorch/utils.py
chMoussa: Hm ok weird indeed, never got this one before.

jpmoutinho: Can you open an issue about this? Just put there a snippet of the previous code, the one you replaced, and the description of the error. Thanks!

tests/test_analog.py
```diff
@@ -388,3 +390,27 @@ def apply_hamevo_and_compare_expected(psi, values):
...
+
+
+@pytest.mark.parametrize(
+    "generator, time_param, result",
+    [
+        (RX(0, "x"), "x", True),
+        (RX(1, 0.5), "y", False),
+        (RX(0, "x"), "y", False),
+        (RX(0, "x"), torch.tensor(0.5), False),
+        (RX(0, torch.tensor(0.5)), torch.tensor(0.5), False),
```
jpmoutinho, line 402: Technically `RX` should not be a generator, I don't think it is Hermitian. For the purpose of this test it's not so relevant, but still I would keep only cases using `Scale` and expand those a little bit.

jpmoutinho (commented):
Thanks @vytautas-a, almost there but I think we should keep the usage of `ConcretizedCallable` as a parameter in the `Parametric` as well for consistency, since you added it to the `Scale`.

pyqtorch/utils.py
vytautas-a: Raised an issue: #283

tests/test_analog.py
vytautas-a: As discussed, I'll leave the `RX` so that test handles all edge cases.

pyqtorch/primitives/parametric.py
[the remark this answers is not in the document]
vytautas-a: restored

pyqtorch/primitives/parametric.py
```diff
@@ -105,13 +105,29 @@ def parse_constant(
...
+            return Parametric._expand_values(self.param_name(values))  # type: ignore [operator]
+
         if param_name == "":
             self.parse_values = parse_tensor
             self.param_name = self.param_name
         elif isinstance(param_name, str):
             self.parse_values = parse_values
         elif isinstance(param_name, (float, int, torch.Tensor)):
             self.parse_values = parse_constant
+        elif isinstance(param_name, ConcretizedCallable):
```
chMoussa, line 129: else instead of elif

pyqtorch/composite/compose.py
```diff
@@ -69,12 +71,14 @@ def forward(
...
-            if isinstance(self.param_name, str)
-            else self.param_name
-        )
-        return scale * self.operations[0].forward(state, values, embedding)
+        if isinstance(self.param_name, str):
+            scale = values[self.param_name]
+        elif isinstance(self.param_name, Tensor):
+            scale = self.param_name
+        elif isinstance(self.param_name, ConcretizedCallable):
+            scale = self.param_name(values)
```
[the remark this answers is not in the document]
vytautas-a: there is this construction without `else` in several places and I think it's kind of a type control. Thus, converting to `else` will defeat this purpose. Maybe add something like
```
...
else:
    raise TypeError(...)
```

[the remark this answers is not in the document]
chMoussa: Sounds good to me also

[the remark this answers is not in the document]
vytautas-a: added a type check in the constructor and left the if .. elif clauses unmodified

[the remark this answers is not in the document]
chMoussa: Ok thanks!

vytautas-a (approved):

Issue #252, opened 2024-08-05
[Noise, Refactor] Make the NoiseProtocol logic accept a sequence of noises


jpmoutinho:
Current the logic for a sequence of noises uses a `dict` of `NoiseProtocol`, as seen in the `QuantumOperation.forward`. As suggested by @Roland-djee [here](https://github.com/pasqal-io/pyqtorch/pull/245#discussion_r1703938325), instead, it would be simpler if the protocol itself contains the sequence.

PR #279, opened 2024-09-03
[Refact] NoiseProtocol with list of noise types

jpmoutinho:
Closes https://github.com/pasqal-io/pyqtorch/issues/252

### DIFF
--- a/pyqtorch/noise/gates.py
+++ b/pyqtorch/noise/gates.py
@@ -17,7 +17,7 @@ class Noise(torch.nn.Module):
         self,
         kraus: list[Tensor],
         target: int,
-        error_probabilities: tuple[float, ...] | float,
+        error_probability: tuple[float, ...] | float,
     ) -> None:
         super().__init__()
         self.target: int = target
@@ -26,10 +26,10 @@ class Noise(torch.nn.Module):
             self.register_buffer(f"kraus_{index}", tensor)
         self._device: torch.device = kraus[0].device
         self._dtype: torch.dtype = kraus[0].dtype
-        self.error_probabilities: tuple[float, ...] | float = error_probabilities
+        self.error_probability: tuple[float, ...] | float = error_probability
 
     def extra_repr(self) -> str:
-        return f"target: {self.qubit_support}, prob: {self.error_probabilities}"
+        return f"target: {self.qubit_support}, prob: {self.error_probability}"
 
     @property
     def kraus_operators(self) -> list[Tensor]:
--- a/pyqtorch/noise/protocol.py
+++ b/pyqtorch/noise/protocol.py
@@ -20,15 +20,15 @@ class NoiseProtocol:
     Defines a protocol of noisy quantum channels.
 
     Args:
-        protocol: single NoiseType instance, list of NoiseType instances or
-            a dict of NoiseType instances. When passing a dict of NoiseType instances,
-            each value should be a dict of options containing the "error_probability",
+        protocol: single NoiseType instance or list of NoiseType instances, or
+            list of (NoiseType, options) tuple. When passing list of tuples,
+            for each NoiseType the options should be a dict containing the "error_probability",
             and optionally a "target". If no "target" is present, the noise instance
             will be applied to the same target of the gate it is used on.
         error_probability: probability of error when passing a single NoiseType
-            or a list of NoiseTypes. Note that all noise types require a single float
-            for the error_probability, while the PauliChannel requires a tuple
-            (px, py, pz) of error probabilities.
+            or a list of NoiseTypes. Note that all noise types require a single float for the
+            error_probability, while the PauliChannel and GeneralizedAmplitudeDamping require
+            a tuple of error probabilities.
 
     Examples:
     ```
@@ -60,13 +60,13 @@ class NoiseProtocol:
     ) -> None:
 
         if isinstance(protocol, NoiseType):
-            self.protocol = [(protocol, {"error_probability": error_probability})]
+            self.protocol = [(protocol, {"error_probability": error_probability})]  # type: ignore [misc]
         elif isinstance(protocol, list) and isinstance(protocol[0], NoiseType):
             self.protocol = [
-                (p, {"error_probability": error_probability}) for p in protocol
+                (p, {"error_probability": error_probability}) for p in protocol  # type: ignore [misc]
             ]
         else:
-            self.protocol = protocol
+            self.protocol = protocol  # type: ignore [assignment]
 
         for _, options in self.protocol:
             err = options.get("error_probability")
@@ -78,19 +78,8 @@ class NoiseProtocol:
 
         self.len = len(self.protocol)
 
-    def __repr__(self) -> str:
-        if self.len == 1:
-            noise, options = self.protocol[0]
-            error_probability = options.get("error_probability")
-            target = options.get("target")
-            if target is not None:
-                return f"{str(noise)}(prob: {error_probability}, target: {target})"
-            else:
-                return f"{str(noise)}(prob: {error_probability})"
-        elif self.len >= 1:
-            return f"NoiseProtocol(length = {self.len})"
-
-    def to_gates(self) -> list:
+    @property
+    def gates(self) -> list:
         gate_list = []
         for noise, options in self.protocol:
             try:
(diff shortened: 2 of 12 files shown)


jpmoutinho:
@EthanObadia did a bit of a refactor in the `NoiseProtocol` to allow a single protocol to hold the info of various noise instances.

Note that for the code that you have running directly on PyQ this breaks the interface. But it should be easy to adapt. Let me know if you have any comments :)

EthanObadia (commented):
@jpmoutinho Thank you very much for the work done in this pull request and for keeping me informed of the progress regarding the noise. I find it much clearer this way, particularly the `forward pass`, which is really well simplified. 

However, I noticed something that doesn't seem to be the intended behavior.

In the following case:

```python
protocol = NoiseProtocol(NoiseType.BITFLIP, error_probability=0.5, target=2)
print(protocol)
```

The output is as expected:

```
BitFlip(prob: 0.5, target: 2)
```

But in the following case:

```python
protocol = NoiseProtocol(
    [NoiseType.BITFLIP, NoiseType.PHASEFLIP],
    error_probability=0.5
)
print(protocol)
```

The output obtained is:

```
NoiseProtocol(length = 2)
```

Which implies results like:

```
_repr_noise(protocol)
```

Returns:

```
', noise: NoiseProtocol(length = 2)'
```

And therefore:

```python
protocol_0 = NoiseProtocol.bitflip(0.2)
protocol_1 = NoiseProtocol.pauli_channel((0.1, 0.2, 0.7))
protocol = NoiseProtocol([protocol_0, protocol_1])
x = X(target=0, noise=protocol)
print(x)
```

Returns:

```
X(target: (0,), noise: NoiseProtocol(length = 2))
```

This does not seem to be the intended behavior.

pyqtorch/noise/gates.py
```diff
@@ -17,7 +17,7 @@ def __init__(
         self,
         kraus: list[Tensor],
         target: int,
-        error_probabilities: tuple[float, ...] | float,
+        error_probability: tuple[float, ...] | float,
```
Roland-djee, line 20: hm since we allow for several error probability values, I'd still keep the plural.

pyqtorch/noise/protocol.py
```diff
@@ -12,43 +14,144 @@ class NoiseProtocol:
...
-    def __repr__(self) -> str:
-        if self.target:
-            return (
-                f"{self.protocol}(prob: {self.error_probability}, "
-                f"target: {self.target})"
-            )
-        return f"{self.protocol}(prob: {self.error_probability})"
+class NoiseProtocol:
+    """
+    Defines a protocol of noisy quantum channels.
```
Roland-djee, line 20:
```suggestion
    Define a protocol of noisy quantum channels.
```

pyqtorch/noise/protocol.py
```diff
@@ -12,43 +14,144 @@ class NoiseProtocol:
...
+            error_probability, while the PauliChannel and GeneralizedAmplitudeDamping require
+            a tuple of error probabilities.
+
+    Examples:
+    ```
+        from pyqtorch.noise import NoiseProtocol, NoiseType
+
+        # Single noise instance
+        protocol = NoiseProtocol(NoiseType.BITFLIP, error_probability = 0.5)
+        protocol = NoiseProtocol.bitflip(error_probability = 0.5)
```
Roland-djee, line 39: Are these two equivalent ?

pyqtorch/noise/protocol.py
```diff
@@ -12,43 +14,144 @@ class NoiseProtocol:
...
+    ) -> None:
+
+        self._error_probability = error_probability
+        self._target = target
+
+        if isinstance(protocol, NoiseType):
+            self.noise_instances = [(protocol, error_probability, target)]
+        elif isinstance(protocol, list) and isinstance(protocol[0], NoiseType):
+            self.noise_instances = [(p, error_probability, target) for p in protocol]  # type: ignore [misc]
+        elif isinstance(protocol, list) and isinstance(protocol[0], NoiseProtocol):
```
Roland-djee, line 69: I would use a predicate:

```python
is_protocol_list = isinstance(protocol, list)
```

and use it in the two clauses as it gets evaluated only once.

pyqtorch/noise/protocol.py
```diff
@@ -12,43 +14,144 @@ class NoiseProtocol:
...
+            self.noise_instances = []
+            for p in protocol:
+                self.noise_instances.extend(p.noise_instances)  # type: ignore [union-attr]
+        else:
+            raise TypeError("Incorrect protocol argument.")
```
Roland-djee, line 74:
```suggestion
            raise TypeError(f"Incorrect argument for protocol {protocol}. Got {protocol[0]}.")
```

pyqtorch/noise/protocol.py
```diff
@@ -12,43 +14,144 @@ class NoiseProtocol:
...
+
+        for _, error_probability, _ in self.noise_instances:
```
Roland-djee, line 76:
```suggestion
        for protocol, error_probability, _ in self.noise_instances:
```

pyqtorch/noise/protocol.py
```diff
@@ -12,43 +14,144 @@ class NoiseProtocol:
...
+            if error_probability is None:
+                raise ValueError(
+                    "This protocol requires a value to the error_probability argument"
```
Roland-djee, line 79:
```suggestion
                    f"Protocol {protocol} requires a non-negative error_probability argument of type `float`. Got {error_probability}."
```

Actually, checking that error probabilities are non-negative might be a good check to add here.

pyqtorch/noise/protocol.py
```diff
@@ -12,43 +14,144 @@ class NoiseProtocol:
...
+        protocol: NoiseType | list[NoiseType] | list[NoiseProtocol],
+        error_probability: tuple[float, ...] | float | None = None,
+        target: int | None = None,
+    ) -> None:
+
+        self._error_probability = error_probability
+        self._target = target
+
+        if isinstance(protocol, NoiseType):
+            self.noise_instances = [(protocol, error_probability, target)]
```
Roland-djee, line 66: Maybe use a small dataclass for these tuples ? So that there is no need to index them later.

pyqtorch/noise/protocol.py
```diff
@@ -12,43 +14,144 @@ class NoiseProtocol:
...
-        return self.options.get("target")
+    def gates(self) -> list:
+        gate_list = []
+        for noise, error_probability, target in self.noise_instances:
+            try:
+                gate_class = getattr(sys.modules["pyqtorch.noise.gates"], str(noise))
+                gate_list.append((gate_class, error_probability, target))
+            except AttributeError:
+                raise ValueError(
+                    f"The protocol {str(noise)} has not been implemented in pyq yet."
```
Roland-djee, line 107:
```suggestion
                    f"The protocol {str(noise)} has not been implemented in PyQTorch yet."
```

jpmoutinho:
Hey @EthanObadia, thanks for the comment :)

In fact that is kind of the intended behaviour, but it was a quick fix. The idea was just that I didn't want to spam a huge string everywhere, so for every protocol that has 2 or more noise instances it just prints the length of the protocol. Maybe you have a better idea...

Roland-djee (commented):
Thanks @jpmoutinho. Looking good. Minor cosmetic comments but the main one is about the use of a small dataclass to access and manipulate noise instances. I think it is generally preferable than indexing.

pyqtorch/noise/gates.py
jpmoutinho: I thought about this, but it is meant to keep everything consistent, since in the noise gates themselves the argument is `error_probability`. Also, while two of them accept several values, most only accept one.

pyqtorch/noise/protocol.py
jpmoutinho: So the test here is a bit different: currently it is just meant to check if the user forgot to pass an `error_probability` (since it is an optional argument in the `__init__`, but only when calling `NoiseProtocol([prot1, prot2])`). Currently the test for the actual contents of the error probabilities is down downstream in the instantiation of the actual noise gates, since a few of them do a different check.

pyqtorch/noise/protocol.py
jpmoutinho: Yep this is a nice idea, I refactored it

Roland-djee (approved):
Nice ! Thanks.

PR #281, opened 2024-09-06
[Bug] Quick fix on SWAP noise and tests to make sure all others are good

jpmoutinho:


Doomsk (approved):
seems good to me! tests are passing

Issue #285, opened 2024-10-16
[Feature] Add Readout and possibly Correlated Readout


chMoussa:
### Describe the feature

Readout is available in Qadence and it would seem best to include it in the available noise models. 

### Would you like to work on this issue?

Yes

PR #286, opened 2024-10-18
[Feature] Readout noise

chMoussa:
Adding noise readout (issue #285). I am thinking initially Readout would be specified at the level of the circuit. I am though unsure if this is a good idea given the current Qadence interface of Readout. I wonder if it makes sense to add it as a noise protocol but it is not compatible with the other noise signatures we have. What do you think @jpmoutinho @RolandMacDoland ?

I also fixed the equations display in `noise.md` and remove the annoying logging with Hamevo (it was printing too much on qadence's side).


jpmoutinho (approved):
Hi @chMoussa, from a quick glance, this design looks good to me, not much to say :)

Issue #287, opened 2024-10-24
[Refactoring] Replace empty lists, and dicts by default None on functions and methods args #590


chMoussa:
Empty lists, tuples, dicts as defaults on functions and methods declarations are prone to changing their default values if the variable in the argument is changed throughout the function/method scope.

chMoussa:
Closes with #288

PR #288, opened 2024-10-24
[Refactor] Changing default dict values to None and fo conversion inside functions

chMoussa:
Solves #287


Doomsk (approved):
brave Charles, thanks!

PR #291, opened 2024-11-08
[Feature] Allow time-dependent duration to be passed at runtime

jpmoutinho:
This MR gives the same interface to the `duration` argument of other PyQ parameters, where you can define it as a string and then pass the actual value in the values dict. This allows to define the duration of the time-dependent integration only at runtime, like so:
```
from pyqtorch import X, HamiltonianEvolution, Scale
from pyqtorch import random_state
import torch

hamiltonian = Scale(X(0), "t")

evo = HamiltonianEvolution(hamiltonian, "t", duration = "duration")

state = random_state(2)

values = {"duration": 0.2}

evo(state, values = values)
```


chMoussa (approved):
LGTM

Issue #274, opened 2024-08-30
[Refact] Control over sparse / dense matrices in the time-dependent solvers


jpmoutinho:
### Describe the feature

Add control for users to pick sparse or dense matrices in the time-dependent solvers. It has some implications on performance.

### Would you like to work on this issue?

None

chMoussa:
@vytautas-a is this still relevant? If not, close

chMoussa:
Closing as solves in #294

PR #294, opened 2024-11-15
[Feature] Add batch dimension to input states for time-dependent solvers

### LINKED ISSUE #292: [Bug] Mesolve checks for initial state
### Short description

When I pass a density matrix of shape (4, 4, 1), the initial checks on mesolve transform it into a (16, 16, 1). When passing jump operators (4,4) the multiplication fail. See test below

### What is the expected result?

_No response_

### What is the actual result?

_No response_

### Steps/Code to reproduce

```
@pytest.mark.parametrize("duration", [torch.rand(1), "duration"])
@pytest.mark.parametrize("input_dm", [True,])
def test_timedependent_with_noise(
    tparam: str,
    param_y: float,
    duration: float | str,
    n_steps: int,
    torch_hamiltonian: Callable,
    hamevo_generator: Sequence,
    sin: tuple,
    sq: tuple,
    input_dm: bool,
) -> None:

    psi_start = random_state(2)
    if input_dm:
        psi_start = density_mat(psi_start)
    dur_val = duration if isinstance(duration, torch.Tensor) else torch.rand(1)

    # simulate with time-dependent solver
    t_points = torch.linspace(0, dur_val[0], n_steps)

    list_ops = Depolarizing(0, error_probability=0.1).tensor(2)
    list_ops = [l.squeeze(-1) for l in list_ops]
    solver = SolverType.DP5_ME
    psi_solver = pyq.mesolve(
        torch_hamiltonian, psi_start, list_ops, t_points, solver
    ).states[-1]

    # simulate with HamiltonianEvolution
    embedding = pyq.Embedding(
        tparam_name=tparam,
        var_to_call={sin[0]: sin[1], sq[0]: sq[1]},
    )
    hamiltonian_evolution = pyq.HamiltonianEvolution(
        generator=hamevo_generator,
        time=tparam,
        duration=duration,
        steps=n_steps,
        solver=solver,
        noise_operators=list_ops,
    )
    values = {"y": param_y, "duration": dur_val}
    psi_hamevo = hamiltonian_evolution(
        state=psi_start, values=values, embedding=embedding
    ).reshape(-1, 1)

    assert torch.allclose(psi_solver, psi_hamevo, rtol=RTOL, atol=ATOL)
```

### Tracebacks (optional)

_No response_

### Environment details (optional)

_No response_

None

vytautas-a:
Add batch dimension to input states for time-dependent solvers. Also introduce flag in config of time-dependent solvers to choose sparse/dense tensors for calculations.

Closes #292 and #274


pyqtorch/time_dependent/integrators/adaptive.py
```diff
@@ -126,9 +126,10 @@ def init_tstep(
         """
 
         sc = self.options.atol + torch.abs(y0) * self.options.rtol
+        f0 = f0.to_dense() if self.options.use_sparse else f0
```
chMoussa, line 129: Hm should it be rather `use_dense` or not `self.options.use_sparse`?

pyqtorch/time_dependent/solvers.py
```diff
@@ -30,7 +30,7 @@ def ode_fun(self, t: float, psi: Tensor) -> Tensor:
         with warnings.catch_warnings():
             # filter-out UserWarning about "Sparse CSR tensor support is in beta state"
             warnings.filterwarnings("ignore", category=UserWarning)
-            res = -1j * self.H(t) @ psi.to_sparse()
+            res = -1j * self.H(t) @ psi.to_sparse() if self.options.use_sparse else psi
```
chMoussa, line 33: Just to make sure, I would use use parenthesis:
-1j * self.H(t) @ `(psi.to_sparse() if self.options.use_sparse else psi)`

tests/test_analog.py
```diff
@@ -292,12 +292,14 @@ def test_hamevo_endianness_cnot() -> None:
     assert torch.allclose(wf_cnot, wf_hamevo, rtol=RTOL, atol=ATOL)
 
 
-@pytest.mark.parametrize("duration", [torch.rand(1), "duration"])
```
chMoussa, line 295: I thought we allowed duration to be passed within the param_values with @jpmoutinho recent MR?

tests/test_analog.py
```diff
@@ -292,12 +292,14 @@ def test_hamevo_endianness_cnot() -> None:
...
+@pytest.mark.parametrize("batch_size", [1, 5])
 @pytest.mark.parametrize("ode_solver", [SolverType.DP5_SE, SolverType.KRYLOV_SE])
 def test_timedependent(
     tparam: str,
     param_y: float,
     duration: float,
+    batch_size: int,
     n_steps: int,
     torch_hamiltonian: Callable,
     hamevo_generator: Sequence,
```
jpmoutinho, line 305: Can you also parametrize over the `use_sparse = True / False`?

tests/test_analog.py
jpmoutinho: Yes it is possible. This parametrization was actually to cover both the case where you pass it as a value, or as a string and then into the values dict. So maybe it should be kept (or refactored, but still covering both).

vytautas-a: yes, deleted that and didn't restore - I'll fix that

tests/test_analog.py
vytautas-a: sure

pyqtorch/time_dependent/solvers.py
vytautas-a: it was actually really needed :)

chMoussa (approved):

PR #296, opened 2024-11-15
[Feature] Correlated readout and extending ReadoutNoise

### LINKED ISSUE #285: [Feature] Add Readout and possibly Correlated Readout

chMoussa:
Closes #285 by adding correlated readout. Also, we extend ReadoutNoise to provide different `error_probability` tensors. Now we can provide a 1D tensor where each qubit has a different probability of corruption. Also we can provide a tensor of shape (n_qubits, 2, 2) that avoids recomputing the confusion_matrix if someone is interested in providing it from data, but we can still create noise matrices for corruption.

### DIFF
--- a/pyqtorch/noise/readout.py
+++ b/pyqtorch/noise/readout.py
@@ -1,13 +1,15 @@
 from __future__ import annotations
 
+from abc import ABC
 from collections import Counter
 from enum import Enum
+from math import log
 
 import torch
 from torch import Tensor
 from torch.distributions import normal, poisson, uniform
 
-from pyqtorch.utils import OrderedCounter
+from pyqtorch.utils import OrderedCounter, sample_multinomial
 
 
 class WhiteNoise(Enum):
@@ -150,7 +152,17 @@ def create_confusion_matrices(noise_matrix: Tensor, error_probability: float) ->
     return torch.stack(confusion_matrices)
 
 
-class ReadoutNoise:
+class ReadoutInterface(ABC):
+    def apply_on_probas(self, batch_probs, n_shots: int) -> Tensor:
+        raise NotImplementedError
+
+    def apply_on_counts(
+        self, counters: list[Counter | OrderedCounter], n_shots
+    ) -> list[Counter]:
+        raise NotImplementedError
+
+
+class ReadoutNoise(ReadoutInterface):
     """Simulate errors when sampling from a circuit.
 
     The model is simple as all bits are considered independent
@@ -191,7 +203,7 @@ class ReadoutNoise:
         self.seed = seed
         self.error_probability = error_probability if error_probability else 0.1
         self.noise_distribution = noise_distribution
-        self.confusion_matrices: Tensor = torch.empty((self.n_qubits, 2, 2))
+        self.confusion_matrix: Tensor = torch.empty((self.n_qubits, 2, 2))
 
     def create_noise_matrix(self, n_shots: int) -> Tensor:
         """Create a noise matrix from a noise distribution.
@@ -213,7 +225,7 @@ class ReadoutNoise:
         confusion_matrices = create_confusion_matrices(
             noise_matrix=noise_matrix, error_probability=self.error_probability
         )
-        self.confusion_matrices = confusion_matrices
+        self.confusion_matrix = confusion_matrices
         return noise_matrix
 
     def apply_on_probas(self, batch_probs: Tensor, n_shots: int = 1000) -> Tensor:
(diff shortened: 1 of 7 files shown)


pyqtorch/noise/readout.py
```diff
@@ -125,16 +105,44 @@ def bs_corruption(
     given a noise matrix.
 
     Args:
-        err_idx: A Boolean array of bit string indices that need to be corrupted.
-        sample: A torch.Tensor of bit strings n_shots x n_qubits.
+        err_idx (Tensor): A Boolean array of bit string indices that need to be corrupted.
```
RolandMacDoland, line 108:
```suggestion
        err_idx (Tensor): A Boolean array of bit string indices to be corrupted.
```

pyqtorch/noise/readout.py
```diff
@@ -150,7 +158,17 @@ def create_confusion_matrices(noise_matrix: Tensor, error_probability: float) ->
     return torch.stack(confusion_matrices)
 
 
-class ReadoutNoise:
+class ReadoutInterface(ABC):
+    def apply_on_probas(self, batch_probs, n_shots: int) -> Tensor:
+        raise NotImplementedError
+
+    def apply_on_counts(
+        self, counters: list[Counter | OrderedCounter], n_shots
```
RolandMacDoland, line 166: A possibility for @singledispatch ?

pyqtorch/noise/readout.py
```diff
@@ -164,34 +182,81 @@ class ReadoutNoise:
...
+              and do not compute the confusion as in the other cases.
         noise_distribution (str, optional): Noise distribution type. Defaults to WhiteNoise.UNIFORM.
-        confusion_matrices (Tensor | None): Confusion matrices that can be used for error miigation.
     """
 
     def __init__(
         self,
         n_qubits: int,
-        error_probability: float | None = None,
+        error_probability: float | Tensor | None = None,
```
RolandMacDoland, line 198: Why don't you default it to 0.1 here ?

RolandMacDoland (commented):
Thanks @chMoussa. Minor comments but I think your `apply_on_proba` and `apply_on_counts` can be grouped as they are semantically very similar but differ on the   parameter type which sounds like a good case for single dispatch.

pyqtorch/noise/readout.py
chMoussa: Changed it

chMoussa:
> Thanks @chMoussa. Minor comments but I think your `apply_on_proba` and `apply_on_counts` can be grouped as they are semantically very similar but differ on the parameter type which sounds like a good case for single dispatch.

Hi @RolandMacDoland , I merged them into an `apply` functions using singledispatchmethod. The only thing is that I am not able to use as type list[Counter]. Let me know what you think.

pyqtorch/noise/readout.py
```diff
@@ -150,7 +157,13 @@ def create_confusion_matrices(noise_matrix: Tensor, error_probability: float) ->
     return torch.stack(confusion_matrices)
 
 
-class ReadoutNoise:
+class ReadoutInterface(ABC):
+    @singledispatchmethod
```
RolandMacDoland, line 161: Nice :)

pyqtorch/noise/readout.py
```diff
@@ -150,7 +157,13 @@ def create_confusion_matrices(noise_matrix: Tensor, error_probability: float) ->
     return torch.stack(confusion_matrices)
 
 
-class ReadoutNoise:
+class ReadoutInterface(ABC):
```
RolandMacDoland, line 160: Docstring ?

pyqtorch/noise/readout.py
```diff
@@ -150,7 +157,13 @@ def create_confusion_matrices(noise_matrix: Tensor, error_probability: float) ->
...
+    @singledispatchmethod
+    def apply(self, input_to_corrupt, n_shots):
```
RolandMacDoland, line 162:
```suggestion
    def apply(self, input, n_shots):
```

But I suspect `input` is a reserved keyword.

pyqtorch/noise/readout.py
```diff
@@ -164,34 +177,80 @@ class ReadoutNoise:
...
             noise_distribution (str, optional): Noise distribution type.
               Defaults to WhiteNoise.UNIFORM.
+            confusion_matrix (Tensor, optional): The confusion matrix if available.
+
         """
         self.n_qubits = n_qubits
         self.seed = seed
-        self.error_probability = error_probability if error_probability else 0.1
+        size_error_probability = (1,)
+        if error_probability is not None:
```
RolandMacDoland, line 217: Is this check still relevant since you've got a default value ?

pyqtorch/noise/readout.py
RolandMacDoland: Maybe simply `inputs` ? Since in both cases you've got some iterable.

RolandMacDoland (approved):
Thanks a lot @chMoussa feels nice and clean. Minor quibbles but I guess it is ready to be merged after you've addressed them.

PR #298, opened 2024-11-28
[Feature] Faster sample method with readout

chMoussa:
Related to #296. 

Using the apply method on the probability vector directly in `QuantumCircuit.sample` leads to faster runtime (example on 10 qubits for 1e6 shots leads to runtime in ms instead of seconds).


RolandMacDoland (approved):
Nice ! Thanks a lot.

PR #308, opened 2025-01-08
[Bugfix] Fix mean when empty tensor for readout

chMoussa:
Solving the edge case when mean of an empty tensor returns nan

### DIFF
--- a/pyqtorch/noise/readout.py
+++ b/pyqtorch/noise/readout.py
@@ -148,7 +148,11 @@ def create_confusion_matrices(noise_matrix: Tensor, error_probability: float) ->
     confusion_matrices = []
     for i in range(noise_matrix.size()[1]):
         column_tensor = noise_matrix[:, i]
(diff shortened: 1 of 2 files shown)


pyqtorch/noise/readout.py
```diff
@@ -148,7 +148,11 @@ def create_confusion_matrices(noise_matrix: Tensor, error_probability: float) ->
     confusion_matrices = []
     for i in range(noise_matrix.size()[1]):
         column_tensor = noise_matrix[:, i]
-        flip_proba = column_tensor[column_tensor < error_probability].mean().item()
+        flip_proba = column_tensor[column_tensor < error_probability]
```
mlahariya, line 151: Seems like a nice place to try using the walrus operator if you are interested. :D

```
fp = fp.mean().item() if len(fp := column_tensor[column_tensor < error_probability]) > 0 else 0.0
```

mlahariya (approved):
Looks good.

PR #310, opened 2025-01-16
[Docs] Delete unnecessary line as coverage should be automated.

RolandMacDoland:
Minor fix to trigger correct coverage.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,7 +38,7 @@ jobs:
             hatch -v -e tests
         - name: Perform unit tests
           run: |
-            hatch -e tests run pytest -n auto
+            hatch -v run test
         - name: Upload coverage reports to Codecov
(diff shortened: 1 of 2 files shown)


mlahariya (approved):
LGTM!

Just to clarify
The default test behaviour `hatch run test` is still `pytest -n auto` ?
The issue was just that there were two seperate `test` and `test-cov` defined in the hatch.envs.tests.scripts?

RolandMacDoland:
> LGTM!
> 
> Just to clarify The default test behaviour `hatch run test` is still `pytest -n auto` ? The issue was just that there were two seperate `test` and `test-cov` defined in the hatch.envs.tests.scripts?

Yeah, essentially the CI workflow didn't call the test command defined in the `pyproject.toml`. So the XML coverage wasn't uploaded to codecov correctly.

PR #312, opened 2025-02-21
[Refactoring] Avoid calling block to tensor on diagonal Hamiltonians

chMoussa:
Solves #175 by:

- [x] : Adding the `diagonal` argument to `QuantumOperation` and `Primitive`, and introducing the `is_diagonal` attribute for detecting such cases where 2D tensor are diagonals with a batchsize, 2D tensors without batchsize but can be diagonal,or 3D tensors are passed and they represent a diagonal operator.
- [x] : Adding `is_diagonal` to other operators such as `Sequence`
- [x] : Adding the `diagonal` argument to tensor methods.
- [x] : Changing utils and matrices methods for diagonal cases. 
- [x] : Dealing with diagonal generators in HamiltonianEvolution
- [x] : Testing diagonal cases


pyqtorch/hamiltonians/evolution.py
```diff
@@ -74,40 +73,29 @@ class GeneratorType(StrEnum):
...
+    if diagonal:
+        evol_operator = torch.transpose(hamiltonian, 0, 1) * (
+            -1j * time_evolution
+        ).view((-1, 1))
+        evol_operator = torch.diag_embed(torch.exp(evol_operator))
+    # for case 3D tensor is passed as generator
+    elif is_diag_batched(hamiltonian):
         evol_operator = torch.diagonal(hamiltonian) * (-1j * time_evolution).view(
             (-1, 1)
         )
```
jpmoutinho, line 101: If I understand correctly you are not yet fixing the batched case?

pyqtorch/hamiltonians/evolution.py
```diff
@@ -222,13 +212,14 @@ def __init__(
                 generator = [generator]
                 self.generator_type = GeneratorType.PARAMETRIC_OPERATION
             else:
+                # avoiding using dense tensor for diagonal generators
+                tgen = generator.tensor(diagonal=generator.is_diagonal)
```
jpmoutinho, line 216: Nice!

jpmoutinho (approved):
Really great work @chMoussa!

Out of curiosity, did you run any benchmarks?

pyqtorch/hamiltonians/evolution.py
chMoussa: It is for the extreme case where you may still pass a 3D-diagonal dense tensor, which may not be differentiated with a 2D case

PR #321, opened 2025-03-25
[Feature] Mutated single-qubit gates

chMoussa:
Solves #165


chMoussa:
> Thanks @chMoussa. I guess this is more addressing #165 right? Besides checking if the performance gain is worth it, I think the two main things to check would be:
> 
> * If there is any penalty in memory consumption (I'm especially worried about lines like `phase_mask = torch.ones_like(state, dtype=state.dtype, device=state.device)`)
> * If all Qadence tests pass with this.

@jpmoutinho I quicky rectified this for the phase mutation. For the Y gate though, a new tensor is necessary but it does not seem to impact the memory usage. For the H gate, I gave up as  this was slower. Qadence tests pass. It does seem we get a factor 2 or more depending on the input size.

pyqtorch/primitives/primitive_gates.py
```diff
@@ -1,35 +1,169 @@
...
+        embedding: Embedding | None = None,
+    ) -> Tensor:
+        values = values or dict()
+        if isinstance(state, DensityMatrix):
+            return super()._forward(state, values, embedding)
+        else:
+            return self._mutate_state_vector(state)
+
+
+class Phase1MutatePrimitive(MutatePrimitive):
```
RolandMacDoland, line 109: Not too sure I get that one.

pyqtorch/primitives/primitive_gates.py
```diff
@@ -1,35 +1,169 @@
...
+    """
+    # Reshape back to original structure
+    state = state.reshape(original_shape)
+
+    # Transpose back to original order
+    inverse_perm = [perm.index(i) for i in range(len(perm))]
+    return state.permute(inverse_perm)
+
+
+class MutatePrimitive(Primitive):
```
RolandMacDoland, line 69:
```suggestion
class MutablePrimitive(Primitive):
```

pyqtorch/primitives/primitive_gates.py
RolandMacDoland, line 109:
```suggestion
class PhaseMutablePrimitive(MutablePrimitive):
```

RolandMacDoland (approved):
Hey @chMoussa Thanks for this. I probably need to understand the logic a bit more for a meaningful review. Otherwise looking good.

pyqtorch/primitives/primitive_gates.py
chMoussa: So these are for the case when, given an operation whose effect is only to multiply the state |1> by a coefficient, apply just a specific modifier caller.

PR #324, opened 2025-04-03
[Feature] Controlled mutable primitives

chMoussa:
Solves #165 for controlled primitives such as Toffoli, CZ and CX

### DIFF
--- a/pyqtorch/primitives/mutation_utils.py
+++ b/pyqtorch/primitives/mutation_utils.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import torch
 from torch import Tensor
 
 
@@ -52,25 +51,17 @@ def mutate_revert_modified(
     return state.permute(inverse_perm)
 
 
-def mutate_control_mask(state: Tensor, control_qubits: tuple[int, ...]) -> Tensor:
-    """Create a mask for mutable operations.
+def mutate_control_slice(n_qubits: int, control_qubits: tuple[int, ...]) -> tuple:
+    """Create a slice for mutable controlled operations.
 
     Args:
-        state (Tensor): Input state
+        n_qubits (int): Number of qubits
         control_qubits (tuple[int, ...]): Control indices.
 
     Returns:
-        Tensor: Mask.
+        tuple: slices to select controlled states.
     """
-    state_indices = torch.arange(state.numel()).view(state.shape)
-
-    # Start with all True and AND with each control qubit condition
-    control_mask = torch.ones_like(state_indices, dtype=torch.bool)
-
+    selector: list = [slice(None)] * n_qubits
     for qubit in control_qubits:
-        # Check if the specific qubit bit is set to 1
-        qubit_is_one = ((state_indices >> qubit) & 1).bool()
-        # AND with our accumulating mask - all must be True
-        control_mask &= qubit_is_one
-
-    return control_mask
+        selector[qubit] = 1
+    return tuple(selector)
--- a/pyqtorch/primitives/mutation_utils.py
+++ b/pyqtorch/primitives/mutation_utils.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import torch
 from torch import Tensor
 
 
@@ -49,3 +50,27 @@ def mutate_revert_modified(
     # Transpose back to original order
     inverse_perm = [perm.index(i) for i in range(len(perm))]
     return state.permute(inverse_perm)
+
+
+def mutate_control_mask(state: Tensor, control_qubits: tuple[int, ...]) -> Tensor:
+    """Create a mask for mutable operations.
+
+    Args:
+        state (Tensor): Input state
+        control_qubits (tuple[int, ...]): Control indices.
+
+    Returns:
+        Tensor: Mask.
+    """
+    state_indices = torch.arange(state.numel()).view(state.shape)
+
+    # Start with all True and AND with each control qubit condition
+    control_mask = torch.ones_like(state_indices, dtype=torch.bool)
+
+    for qubit in control_qubits:
+        # Check if the specific qubit bit is set to 1
+        qubit_is_one = ((state_indices >> qubit) & 1).bool()
+        # AND with our accumulating mask - all must be True
+        control_mask &= qubit_is_one
+
+    return control_mask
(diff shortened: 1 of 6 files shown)


pyqtorch/primitives/mutation_utils.py
```diff
@@ -49,3 +49,19 @@ def mutate_revert_modified(
...
+    Args:
+        n_qubits (int): Number of qubits
+        control_qubits (tuple[int, ...]): Control indices.
+
+    Returns:
+        tuple: slices to select controlled states.
+    """
+    selector: list = [slice(None)] * n_qubits
+    for qubit in control_qubits:
+        selector[qubit] = 1
```
RolandMacDoland, line 66: Do I understand correctly that this will end up being a list of None initialized slice objects but for the controlled indices being 1 ?

pyqtorch/primitives/primitive.py
```diff
@@ -163,3 +166,58 @@ def apply_phase1(self, state) -> Tensor:
...
+    ):
+        super().__init__(operation, control, target, noise=noise)
+        self.modifier: Callable = modifier
+
+    def _mutate_state_vector(self, state: Tensor) -> Tensor:
+        result = state.clone()
+
+        # first, create a mask to separate states to be modified
+        mask = mutate_control_slice(len(state.shape) - 1, self.control)
+        controlled_state = state[mask]
```
RolandMacDoland, line 197: Do I understand correctly that this selects the subspace of control states ?

pyqtorch/primitives/primitive.py
```diff
@@ -163,3 +166,58 @@ def apply_phase1(self, state) -> Tensor:
...
+        # second, modify only the controlled states
+        result[mask] = self._controlled_mutate(controlled_state)
```
RolandMacDoland, line 199:
```suggestion
        # second, mutate the control states only
        result[mask] = self._mutate_control_state(control_state)
```

pyqtorch/primitives/primitive.py
```diff
@@ -163,3 +166,58 @@ def apply_phase1(self, state) -> Tensor:
...
+        return result
+
+    def _controlled_mutate(self, controlled_state: Tensor) -> Tensor:
+        controlled_state_shape = controlled_state.shape
+        perm, controlled_state = mutate_separate_target(controlled_state, self.target)
+
+        controlled_state = self.modifier(controlled_state)
+        controlled_state = mutate_revert_modified(
+            controlled_state, controlled_state_shape, perm
+        )
```
RolandMacDoland, line 210: Not sure I totally follow but let's see:
1. You generate a permutation matrix for the control states
2. You apply the gate to the control states only
3. You revert the permutation back to the original state 
Is it what I should understand ?

pyqtorch/primitives/primitive_gates.py
```diff
@@ -201,6 +203,14 @@ def __init__(
     ):
         super().__init__("X", control, target, noise=noise)
 
+    def _controlled_mutate(self, controlled_state: Tensor) -> Tensor:
```
RolandMacDoland, line 206: Not too sure what this does but I can get that you flip the qubits for which indices are lower than the target qubit one ?

pyqtorch/primitives/primitive_gates.py
```diff
@@ -224,6 +234,13 @@ def __init__(
     ):
         super().__init__("Z", control, target, noise=noise)
 
+    def _mutate_state_vector(self, state):
+        result = state.clone()
+        all_controls = self.control + self.target
```
RolandMacDoland, line 239:
```suggestion
        all_qubits = self.control + self.target
```

tests/primitives/test_mutation.py
```diff
@@ -0,0 +1,66 @@
...
+    gate = op(target)
+
+    primitive_op = Primitive(gate.operation, qubit_support=gate.qubit_support)
+    assert torch.allclose(gate(state), primitive_op(state), atol=ATOL)
+
+
+@pytest.mark.parametrize(
+    "op, op_str",
+    [
+        (pyq.CNOT, "X"),
```
RolandMacDoland, line 42: Don't we have the Toffoli gate in PyQTorch ? More generally multi-controlled unitaries ?

RolandMacDoland (commented):
Hey @chMoussa thanks a lot ! Looks nice but I'd like to get a bit more insights for a meaningful review.

pyqtorch/primitives/mutation_utils.py
chMoussa: yes

pyqtorch/primitives/primitive.py
chMoussa: yes

chMoussa: Or more like, selecting the states that comply with the controls

tests/primitives/test_mutation.py
chMoussa: yes, I should try this as next step.

RolandMacDoland: Ah OK. Do you plan to include it in this PR ?

pyqtorch/primitives/primitive.py
RolandMacDoland: Yeah but what I understand is that all other elements in `mask` but the control states are `slice(None)`. I just fail to visualise what `controlled_state` looks like. It feels it has the same dimensionality as `state` but might be wrong there.

pyqtorch/primitives/primitive.py
RolandMacDoland: Yeah, maybe `modifier` is a bit misleading since it feels like it is an identity lambda from the default.

chMoussa: Yes, not sure what to set as default.

pyqtorch/primitives/primitive.py
chMoussa: We remove the control qubit dimension from the resulting tensor. So you go from a shape `(2,...,2)` of `n_qubits` to a `n_qubits - len(controls)` shape

pyqtorch/primitives/primitive_gates.py
chMoussa: As we have selected the control_states, we need to shift indices accordingly

tests/primitives/test_mutation.py
chMoussa: Yes, it's already here

RolandMacDoland (approved):
Cool, thanks @chMoussa. LGTM.

PR #326, opened 2025-04-07
[Feature] Differentiation with AD on observable parameters only

### LINKED ISSUE #325: [Feature] Add possibility to separate observable parameters when differentiating with AD
### Describe the feature

Introduce a `values_observable` argument to api functions and be able to autograd with respect to it. Note we should raise an error for GPSR and Adjoint differentiation as one cannot obtain derivatives with these protocols.

### It should be implemented because

_No response_

### Additional context

_No response_

None

chMoussa:
- [x] Add a `values_observable` argument in api function
- [x] Raise errors for other differentiation methods
- [x] Test on a simple example

### DIFF
--- a/pyqtorch/api.py
+++ b/pyqtorch/api.py
@@ -171,7 +171,9 @@ def sampled_expectation(
     # batchsize needs to be first dim for eigh
     eigvals, eigvecs = torch.linalg.eigh(
         observable.tensor(
-            values=values_observable, embedding=embedding, full_support=tuple(range(n_qubits))
+            values=values_observable,
+            embedding=embedding,
+            full_support=tuple(range(n_qubits)),
         ).permute((2, 0, 1))
     )
     eigvals = eigvals.squeeze()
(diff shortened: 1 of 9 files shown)


jpmoutinho (approved):
Thanks @chMoussa, not much to say.

One thing to consider is if, given the intent of making the values interface more complex, there could be value in coming up with a specific values structure to handle that. E.g., maybe a single `dataclass` instead of multiple arguments for dictionaries. That could potentially just be in Qadence though. Or is the idea that in Qadence the values dict will remain unchanged and then the separation is dealt with automatically when passing to the backend?

chMoussa:
> Thanks @chMoussa, not much to say.
> 
> One thing to consider is if, given the intent of making the values interface more complex, there could be value in coming up with a specific values structure to handle that. E.g., maybe a single `dataclass` instead of multiple arguments for dictionaries. That could potentially just be in Qadence though. Or is the idea that in Qadence the values dict will remain unchanged and then the separation is dealt with automatically when passing to the backend?

@jpmoutinho I would probably come with something up in qadence via the backend interface to separate the values given the observable. I'll see how it plays.

Issue #177, opened 2024-05-16
[Feature, Performance] Optimize HamEvo for commuting Hamiltonians


jpmoutinho:
Moving the original issue from Qadence: https://github.com/pasqal-io/qadence/issues/346

With the reorganization of HamEvo, this would make sense to implement in PyQTorch instead.

This is essentially equivalent to doing exact Trotterization automatically whenever possible. Should be properly planned in that context as well, since we have Trotterization code in Qadence.

RolandMacDoland:
@chMoussa would it make sense to pick that one up for next sprint ?

chMoussa:
> @chMoussa would it make sense to pick that one up for next sprint ?

@RolandMacDoland sure I can try.

RolandMacDoland:
> > [@chMoussa](https://github.com/chMoussa) would it make sense to pick that one up for next sprint ?
> 
> [@RolandMacDoland](https://github.com/RolandMacDoland) sure I can try.

Thanks 🙏

PR #328, opened 2025-04-29
[Feature, Performance] Optimize HamiltonianEvolution for commuting hamiltonians

chMoussa:
Solves #177 by:

- [x] Adding checks for commuting terms in the `Add` operator, so cases with different supports and commuting pauli-strings
- [x] Handling the `Add` case with commuting terms when initialising `HamiltonianEvolution`
- [x] Using a `Sequence` of `HamiltonianEvolution` from commuting terms, and modifying `HamiltonianEvolution` methods such as `forward`, `tensor`
- [x] Add commuting cases in tests

### DIFF
--- a/pyqtorch/composite/compose.py
+++ b/pyqtorch/composite/compose.py
@@ -244,7 +244,7 @@ class Add(Sequence):
             return False
         qubit_supports_intersect = [set(op.qubit_support) for op in self.operations]
 
-        disjoint_sets = set()
+        disjoint_sets: set[int] = set()
         for s in qubit_supports_intersect:
             if s.isdisjoint(disjoint_sets):
                 disjoint_sets.update(s)
--- a/pyqtorch/composite/compose.py
+++ b/pyqtorch/composite/compose.py
@@ -217,25 +217,6 @@ class Add(Sequence):
         )
 
     @property
-    def is_pauli_add(self) -> bool:
-        """Check if Add contains only pauli string, that is a weighted
-        sum of tensor products of pauli operators.
-
-        Returns:
-            bool: True if pauli string.
-        """
-        all_leaves_ops: list = list()
-        for op in self.flatten():
-            if isinstance(op, Scale):
-                if isinstance(op.operations[0], Sequence):
-                    all_leaves_ops += op.operations[0].flatten()
-                else:
-                    all_leaves_ops += [op.operations[0]]
-            else:
-                all_leaves_ops.append(op)
-        return all(isinstance(op, PAULI_OPS) for op in all_leaves_ops)  # type: ignore[arg-type]
-
-    @property
     def commuting_terms(self) -> bool:
         """Check if operator can be decomposed into commuting terms."""
 
--- a/pyqtorch/composite/compose.py
+++ b/pyqtorch/composite/compose.py
@@ -144,6 +144,24 @@ class Scale(Sequence):
         return self
 
 
+def paulis_to_binary_vectors(p_string: list[str]) -> tuple[list[int], list[int]]:
+    """Convert a pauli string to a binary representation for symplectic inner product."""
+    x_vec = [1 if p in ["X", "Y"] else 0 for p in p_string]
+    z_vec = [1 if p in ["Z", "Y"] else 0 for p in p_string]
+    return x_vec, z_vec
+
+
+def symplectic_inner_product(
+    x1: list[int], z1: list[int], x2: list[int], z2: list[int]
+) -> int:
+    """Sympletic inner prodtcfor checking commutation fot two pauli strings p1 and p2
+    represented by their binary representation (`x1`, `z1`), (`x2, `z2``)"""
+    product = 0
+    for k in range(len(x1)):
+        product ^= (x1[k] & z2[k]) ^ (z1[k] & x2[k])
+    return product
+
+
 class Add(Sequence):
     """
     The 'add' operation applies all 'operations' to 'state' and returns the sum of states.
@@ -219,6 +237,10 @@ class Add(Sequence):
     def _symplectic_commute(self) -> bool:
         """Check if operator is composed of pauli strings, each commuting with each other.
         Done by computing the symplectic inner product from the string-based representations.
+
+        Reference:
+            S. Aaronson, D. Gottesman, Improved Simulation of Stabilizer Circuits,
+            Phys. Rev. A 70, 052328 (2004). arXiv:quant-ph/0406196
         """
 
         if all(isinstance(op, PAULI_OPS) for op in self._flatten()):
@@ -259,29 +281,17 @@ class Add(Sequence):
                         return False
 
             # Convert to binary vectors
-            binary_vectors = []
-            for p_string in pauli_strings:
-                x_vec = [1 if p in ["X", "Y"] else 0 for p in p_string]
-                z_vec = [1 if p in ["Z", "Y"] else 0 for p in p_string]
-                binary_vectors.append((x_vec, z_vec))  # Keep x and z vectors
+            binary_vectors = [
+                paulis_to_binary_vectors(p_string) for p_string in pauli_strings
+            ]
 
             # Check commutation using symplectic inner product
             n_strings = len(pauli_strings)
             for i in range(n_strings):
                 for j in range(i + 1, n_strings):
-                    vec_i = binary_vectors[i]
-                    vec_j = binary_vectors[j]
-
-                    # Split into x and z parts
-                    x_i = vec_i[0]
-                    z_i = vec_i[1]
-                    x_j = vec_j[0]
-                    z_j = vec_j[1]
-
-                    # Symplectic product: (x_i · z_j) XOR (z_i · x_j)
-                    product = 0
-                    for k in range(nb_support):
-                        product ^= (x_i[k] & z_j[k]) ^ (z_i[k] & x_j[k])
+                    product = symplectic_inner_product(
+                        *binary_vectors[i], *binary_vectors[j]
+                    )
 
                     if product == 1:
                         return False  # Found a pair that doesn't commute
@@ -314,7 +324,10 @@ class Add(Sequence):
         if len(self.operations) == 1:
             return False
 
-        if self._symplectic_commute() or self._disjoint_supports():
+        if self._disjoint_supports():
+            return True
+
+        if self._symplectic_commute():
             return True
 
         return False
--- a/pyqtorch/composite/compose.py
+++ b/pyqtorch/composite/compose.py
@@ -217,8 +217,8 @@ class Add(Sequence):
         )
 
     @property
-    def is_pauli_string(self) -> bool:
-        """Check if Add is a pauli string, that is a weighted
+    def is_pauli_add(self) -> bool:
+        """Check if Add contains only pauli string, that is a weighted
         sum of tensor products of pauli operators.
 
         Returns:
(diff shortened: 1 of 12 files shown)


jpmoutinho (commented):
Thanks @chMoussa, look like a really nice solution. Two comments:
- It's probably worth explicitly using the commuting terms True / False in all tests that use the `random_pauli_hamiltonian`. Also it might be worth explicitly running the Qadence pipeline with this change.
- Did you observe the expected performance gains from the original qadence issue? https://github.com/pasqal-io/qadence/issues/346
-

pyqtorch/hamiltonians/evolution.py
```diff
@@ -345,7 +375,10 @@ def _symbolic_generator(
         return hamiltonian
 
     def _tensor_generator(
```
RolandMacDoland, line 377:
```suggestion
    def _generator(
```

pyqtorch/hamiltonians/evolution.py
```diff
@@ -394,24 +438,82 @@ def spectral_gap(self) -> Tensor:
...
+    def _time_evolution(
+        self, values: dict[str, Tensor] | ParameterDict | None = None
+    ) -> Tensor:
+        """
+        Get the evolution from parameter values.
+        Arguments:
+            values: Values of parameters.
+        Returns:
+            The time evolution.
+        """
```
RolandMacDoland, line 450:
```suggestion
        """Get the evolution from parameter values.
        
        Arguments:
            values: Values of parameters.
            
        Returns:
            The time evolution.
        """
```

pyqtorch/hamiltonians/evolution.py
```diff
@@ -394,24 +438,82 @@ def spectral_gap(self) -> Tensor:
...
+        values = values or dict()
```
RolandMacDoland, line 451: Why not defaulting `values` to `dict()` and not bother with this check ?

pyqtorch/hamiltonians/evolution.py
```diff
@@ -394,24 +438,82 @@ def spectral_gap(self) -> Tensor:
...
+            time_evolution = self.time
+        return time_evolution
+
     def _forward(
         self,
         state: Tensor,
         values: dict[str, Tensor] | ParameterDict | None = None,
         embedding: Embedding | None = None,
     ) -> State:
         values = values or dict()
```
RolandMacDoland, line 470: Yeah I would suggest to default it as early as possible so that we don't repeat that check further down the abstraction.

pyqtorch/hamiltonians/evolution.py
```diff
@@ -394,24 +438,82 @@ def spectral_gap(self) -> Tensor:
...
+        else:
+            return self._forward_commuting_generators(state, values, embedding)
+
+    def _forward_commuting_generators(
+        self,
+        state: Tensor,
+        values: dict[str, Tensor] | ParameterDict | None = None,
+        embedding: Embedding | None = None,
+    ) -> State:
+        """
```
RolandMacDoland, line 486: Docstring needs formatting.

pyqtorch/primitives/primitive_gates.py
```diff
@@ -33,6 +33,7 @@ def __init__(
             target,
             noise=noise,
         )
+        self.name = "X"
```
RolandMacDoland, line 36: Maybe these names should be passed to `super()`.

pyqtorch/composite/compose.py
```diff
@@ -215,6 +216,109 @@ def tensor(
...
+            if s.isdisjoint(disjoint_sets):
+                disjoint_sets.update(s)
+            else:
+                return False
+        if len(disjoint_sets) == len(self.qubit_support):
+            return True
+
+        return False
+
+    @property
```
RolandMacDoland, line 308: Not too sure this should be a `@property`. If this is a regular predicate method, it conveys the idea that you compute smtg, which you actually do here.

RolandMacDoland (commented):
Thanks @chMoussa LGTM. Some comments but happy to merge once resolved.

pyqtorch/composite/compose.py
```diff
@@ -215,6 +216,109 @@ def tensor(
             mat,
         )
 
+    def _symplectic_commute(self) -> bool:
```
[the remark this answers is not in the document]
chMoussa: Done

pyqtorch/composite/compose.py
```diff
@@ -215,6 +216,109 @@ def tensor(
...
+                            == ""
+                        ):
+                            pauli_strings[ind_op][
+                                qubit_support.index(subop.qubit_support[0])
+                            ] = subop.name
+                        else:
+                            # cases where two or more operators
+                            # are defined for a same qubit
+                            # are not considered
+                            return False
```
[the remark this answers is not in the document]
chMoussa: yes

pyqtorch/composite/compose.py
```diff
@@ -215,6 +216,109 @@ def tensor(
...
+    @property
+    def commuting_terms(self) -> bool:
+        """Return if all operations are commuting terms.
+        Useful for HamiltonianEvolution.
+        """
+
+        if len(self.operations) == 1:
+            return False
+
+        if self._symplectic_commute() or self._disjoint_supports():
```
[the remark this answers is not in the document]
chMoussa: Hm could be but I thought this edge case could be treated separately to avoid going through doing the instructions of disjoint_supports

pyqtorch/hamiltonians/evolution.py
```diff
@@ -291,9 +317,12 @@ def is_parametric_generator(self) -> bool:
     def is_time_dependent(self) -> bool:
         return self._is_time_dependent(self.generator)
 
-    def flatten(self) -> ModuleList:
+    def _flatten(self) -> ModuleList:
```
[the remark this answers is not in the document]
chMoussa: As Hamevo is a sequence, to not support it for pauli-commutation checks in Add, I added this

pyqtorch/hamiltonians/evolution.py
chMoussa: Even better, removing this

pyqtorch/composite/compose.py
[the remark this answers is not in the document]
RolandMacDoland: Yeah, but I feel natural that `disjoint_supports` returns `False` when there is only one operation. Then, avoid the check above.

pyqtorch/composite/compose.py
```diff
@@ -215,6 +216,109 @@ def tensor(
...
+                    for subop in op._flatten():
+                        if (
+                            pauli_strings[ind_op][
+                                qubit_support.index(subop.qubit_support[0])
+                            ]
+                            == ""
+                        ):
+                            pauli_strings[ind_op][
+                                qubit_support.index(subop.qubit_support[0])
+                            ] = subop.name
```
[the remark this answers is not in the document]
chMoussa: No need anymore

pyqtorch/composite/compose.py
```diff
@@ -215,6 +216,109 @@ def tensor(
...
+                    # Split into x and z parts
+                    x_i = vec_i[0]
+                    z_i = vec_i[1]
+                    x_j = vec_j[0]
+                    z_j = vec_j[1]
+
+                    # Symplectic product: (x_i · z_j) XOR (z_i · x_j)
+                    product = 0
+                    for k in range(nb_support):
+                        product ^= (x_i[k] & z_j[k]) ^ (z_i[k] & x_j[k])
```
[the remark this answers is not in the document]
chMoussa: Yes, changed to using now numpy arrays of integers for the symplectic product

pyqtorch/primitives/primitive_gates.py
chMoussa: Removed them eventually

pyqtorch/composite/compose.py
```diff
@@ -215,6 +216,109 @@ def tensor(
...
+                                qubit_support.index(subop.qubit_support[0])
+                            ] = subop.name
+                        else:
+                            # cases where two or more operators
+                            # are defined for a same qubit
+                            # are not considered
+                            return False
+                else:
+                    if (
+                        pauli_strings[ind_op][qubit_support.index(op.qubit_support[0])]
```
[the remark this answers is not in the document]
chMoussa: No need anymore after refactor

pyqtorch/composite/compose.py
```diff
@@ -215,6 +216,109 @@ def tensor(
...
+        """Check if operator is composed of pauli strings, each commuting with each other.
+        Done by computing the symplectic inner product from the string-based representations.
+        """
+
+        if all(isinstance(op, PAULI_OPS) for op in self._flatten()):
+            qubit_support = self.qubit_support
+            nb_support = len(qubit_support)
+
+            # build string representation of pauli operations
+            pauli_strings = [[""] * nb_support for _ in range(len(self.operations))]
```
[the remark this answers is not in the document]
chMoussa: Refactored to avoid strings

chMoussa:
> Thanks @chMoussa, look like a really nice solution. Two comments:
> 
> * It's probably worth explicitly using the commuting terms True / False in all tests that use the `random_pauli_hamiltonian`. Also it might be worth explicitly running the Qadence pipeline with this change.
> * Did you observe the expected performance gains from the original qadence issue? [[Feature, Performance] Optimize HamEvo for commuting Hamiltonians qadence#346](https://github.com/pasqal-io/qadence/issues/346)

@jpmoutinho Yes, just added this. I did witness significant gains!

chMoussa:
> Thanks @chMoussa LGTM. Some comments but happy to merge once resolved.

Hey @RolandMacDoland , thanks for the feedback! We generate now directly the binary representations instead of using strings.

jpmoutinho (approved):
Great, all good from my side.

pyqtorch/hamiltonians/evolution.py
```diff
@@ -172,6 +172,8 @@ def __init__(
         self.duration = duration
         self.use_sparse = use_sparse
         self.is_diagonal = False
+        self._param_uuid = str(uuid4())
+        original_generator = None
```
RolandMacDoland, line 176: What is it for exactly ?

pyqtorch/hamiltonians/evolution.py
```diff
@@ -206,6 +208,33 @@ def __init__(
             self.generator_type = GeneratorType.SYMBOL
             self.generator_symbol = generator
             generator = []
+        elif (
+            noise is None and isinstance(generator, Add) and generator.commuting_terms()
+        ):
+            qubit_support = generator.qubit_support
+            self.generator_type = (
```
RolandMacDoland, line 215: The more I see this machinery to define generators, the more it feels like it needs a custom type to handle it cleanly. Probably in another PR.

RolandMacDoland (commented):
Thanks @chMoussa. Few more comments but LGTM. What I'd suggest in further PRs: 1) default values to empty dict as soon as possible to avoid redundant checks all over the place 2) think about a custom Generator type.

pyqtorch/hamiltonians/evolution.py
chMoussa: it is used in _generator, as for the commuting case, we store a sequence of Hamevo, and original_generator makes us directly get the tensor of the hamiltonian without storing a single primitive

pyqtorch/hamiltonians/evolution.py
chMoussa: @RolandMacDoland I added two extra types then, one for parametric commuting, the other for non parametric commuting cases.

PR #331, opened 2025-05-16
[BugFix] Fix recursive gpsr

### LINKED ISSUE #330: [Bug] GPSR with repeated parameter cannot be differentiated twice with shots
### Short description

With the current implementation, second order derivative do not work when asking for shots. The context used in the GPSR differentiation method fails to save the right values for the backward pass.

### What is the expected result?

_No response_

### What is the actual result?

_No response_

### Steps/Code to reproduce

The code from the DQC docs can be used but using a positive n_shots to reproduce the error.

### Tracebacks (optional)

_No response_

### Environment details (optional)

_No response_

Yes

chMoussa:
Re-allowing to use `PSRExpectation.apply` to be able to backpropagate many times with PSR. Solves #330


mlahariya (approved):
LGTM!

Issue #329, opened 2025-05-14
[Feature] Add two-qubit digital noise


chMoussa:
### Describe the feature

So far, we have available single-qubit noise operations. Let us extend to two-qubit operations with:

- Two-qubit depolarizing channel: described by mapping an input a density matrix $\rho$ to the Kraus representation $\sum_i p_i  K_i \rho K_i^{\dagger}$ where the $K_i$ are the 16 two-qubit pauli matrices. Let us implement a case where we reduce to a single-input probability $p$, for which the formula becomes: 
$$(1-p) \rho + \frac{p}{15} (IX \rho IX + IY \rho IY + IZ \rho IZ + XI \rho XI + ... + ZZ \rho ZZ) $$
- Two-qubit dephasing channel: with a Kraus representation: 
$$(1-p) \rho + \frac{1}{3} (IZ \rho IZ + ZI \rho ZI + ZZ \rho ZZ)$$

### It should be implemented because

We only support single-qubit noise operations.

### Additional context

They should be implemented following the `pyqtorch/noise/digital_gates.py` module. Furthermore, `pyqtorch/noise/digital_protocol.py` module should be extended to accept two-qubit operations. So adding two `DigitalNoiseType` values, the `target` attribute of the `DigitalNoiseInstance` and `DigitalNoiseProtocol` instance should be extended to accept tuples of two integers.
Regarding `DigitalNoiseProtocol`, two new methods should be added for the two new operations. 

Also, when initializing a `QuantumOperation`, we should check the consistency of the input `noise`, e.g. that the qubit the operation and the noise match.

Finally, we should add tests in `tests/noise/test_noise.py`.

### Would you like to work on this issue?

None

Maxtax92:
Assign to me

chMoussa:
> Assign to me

@Maxtax92 Actually, the workflow for uhack is that you first need to open a MR, and when approved/merged, I assign the issue.

PR #334, opened 2025-05-28
[UnitaryHack] Add two-qubit digital noise

AprilSweettooth:


AprilSweettooth:
Hey @chMoussa , I am sorry I never used hatch before. I just locally ran pytest on this, and it passed. Shall I try hatch?

chMoussa:
> Hey @chMoussa , I am sorry I never used hatch before. I just locally ran pytest on this, and it passed. Shall I try hatch or is there a difference?

When I run CI on your MR, many noise tests fails. Do you see the CI workflow results? 

Also, I suggest first to use pre-commit to apply linting and reformatting of files. 

For hatch, no need to install it but it helps avoiding to run the long pytest command defined in the pyproject.toml.
Once you install it, you can:
# enter default env
hatch shell
# run all tests as 
hatch -e tests run test

AprilSweettooth:
Hey @chMoussa , thanks for the advice! So now I have applied pre-commit. I don't know if it is going to show up, but there are a few errors in the test that are not related to my new implementation (two-qubit noise). Not sure how to get around them.

chMoussa (changes requested):
Hey @AprilSweettooth, basically you need to modify also the current tests, even if they are not part of your implementation. See my feedback.

tests/noise/test_noise.py
```diff
@@ -384,3 +385,48 @@ def test_analog_noise_add():
         noise_add.noise_operators
     )
     assert noise_add.qubit_support == (2, 3)
+
+
+def test_two_qubit_depolarizing():
```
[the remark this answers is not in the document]
AprilSweettooth: Hey @chMoussa , thanks again for the useful suggestions. After the modification, I still got just one error during test, saying ' tests/time_dependent/test_solvers.py::test_sesolve[dp5_se-1] - RuntimeError: Numpy is not available'. Is it a local thing, or is something still wrong with my current implementation? Thanks!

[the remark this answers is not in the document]
chMoussa: Could you pass the full error message/ Logs?

pyqtorch/noise/digital_gates.py
```diff
@@ -375,3 +375,66 @@ def __init__(
...
+        if error_probability < 0.0 or error_probability > 1.0:
+            raise ValueError("Error probability must be between 0 and 1.")
+
+        X, Y, Z = XMAT, YMAT, ZMAT
+        paulis = [IMAT, X, Y, Z]
+        P_list = [
+            torch.kron(A, B)
+            for A in paulis
+            for B in paulis
+            if not (torch.allclose(A, IMAT) and torch.allclose(B, IMAT))
```
chMoussa, line 405: I think you could remove the if statement to still include the `IMAT` kron `IMAT` but when assigning the probabilities, you just index well as `sqrt(1.0 - error_probability)` for index `0` and `sqrt(error_probability / 15.0)`  for the rest. That seems more efficient to me :)

pyqtorch/utils.py
```diff
@@ -519,40 +519,72 @@ def expand_operator(
...
+        target_list = list(target)
+    else:
+        raise TypeError("Target must be an int or a tuple of two ints.")
+
+    if any(t >= n_qubits or t < 0 for t in target_list):
+        raise ValueError("Target indices must be within [0, n_qubits).")
+
+    target_list_sorted = sorted(target_list)
+
+    full_op: Union[Tensor, None] = None
```
chMoussa, line 555: Do you need to initialize it to `None` and test for it in the loop? Could simply initializing to `1` works?

tests/noise/test_noise.py
```diff
@@ -308,7 +312,18 @@ def test_digital_noise_apply(
...
+    Goes through all parametric gates and tests their application to a random state
+    in comparison with the noisy version with error_probability = 0.
+    Skips two-qubit noise types (tested separately).
     """
+    # Skip for two-qubit noise types
+    if noise_type in {
+        DigitalNoiseType.TWO_QUBIT_DEPHASING,
+        DigitalNoiseType.TWO_QUBIT_DEPOLARIZING,
+    }:
+        pytest.skip(f"{noise_type.name} is a two-qubit noise type. Skipped.")
```
chMoussa, line 325: Why skipping these cases? I think you can extend the tests to these cases. Try to understand what they are testing for, and adapt them to test the correctness of your new classes.

tests/noise/test_noise.py
```diff
@@ -342,7 +357,18 @@ def test_param_noise_apply(
...
```
chMoussa, line 370: Same.

chMoussa, line 370: Same.

tests/noise/test_noise.py
```diff
@@ -384,3 +410,35 @@ def test_analog_noise_add():
...
+    target = (0, 1)
+    prob = 0.3
+    # Start from a pure product state |00⟩
+    rho_0 = density_mat(product_state("00", batch_size))
+
+    # Apply TwoQubit noise
+    if noise_type == DigitalNoiseType.TWO_QUBIT_DEPOLARIZING:
+        noise = TwoQubitDepolarizing(target=target, error_probability=prob)
+    elif noise_type == DigitalNoiseType.TWO_QUBIT_DEPHASING:
+        noise = TwoQubitDephasing(target=target, error_probability=prob)
```
chMoussa, line 433: You could directly instantiate a `DigitalNoiseProtocol(noise_type, error_probability)` to avoid the if statement.

chMoussa (changes requested):
Hello @AprilSweettooth, going there but I would ask to modify your tests accordingly with the new two-qubit cases in `test_digital_noise_apply` and `test_param_noise_apply`? Simply testing that you have a two-qubit noise, and the op from `OPS_PARAM` has a `qubit_support` of length 2. 
Btw, you have an error appearing in the docs but not the tests, you can see this with `hatch run docs:serve ` to generate locally the docs.

Maybe also to help, on the `promote_operator` function you modified:
You can use the `expand_operator` utils function, and keep `promote_operator` as it was only for the single-qubit noise cases (it would be more performant than `expand_operator`). So an `if` statement in the `tensor` method for the `Noise` class would do.

AprilSweettooth:
Hey @chMoussa , thanks for the suggestions. I've tried to reimplement them with some success. However, I am unable to circumvent the numpy error with time-dependent tests, as well as the documentation errors... Any advice on this matter? Appreciate it! Also, would you check if there are any other errors from your end?

chMoussa:
> Hey @chMoussa , thanks for the suggestions. I've tried to reimplement them with some success. However, I am unable to circumvent the numpy error with time-dependent tests, as well as the documentation errors... Any advice on this matter? Appreciate it! Also, would you check if there are any other errors from your end?

Have you tried with a fresh environment?
try in your local repo:
```
hatch env prune
hatch shell
```
Once the env is loaded, check if `numpy` is installed by entering a `python` shell then try` import numpy `
Now when testing, do:
`hatch -e tests run test
`

chMoussa (changes requested):
Hello @AprilSweettooth, I made a suggestion on the tests. Not sure why you removed your correctness test. I think it is necessary to perform it. Also would be nice to test that an error is thrown when you try setting a noisy single-qubit gate with a two-qubit noise setting.

Also with your current solution, a test fails:
`ERROR tests/noise/test_noise.py::test_noisy_primitive[n_qubits0] - ValueError: The protocol two_qubit_depolarizing has not been implemented in PyQTorch yet.`

tests/noise/test_noise.py
```diff
@@ -356,16 +371,25 @@ def test_param_noise_apply(
...
+        for op in OPS_PARAM:
+            supp = get_op_support(op, n_qubits)
+            params = [f"th{i}" for i in range(op.n_params)]
+            op_concrete = op(*supp, *params)
+            op_concrete_noise = op(*supp, *params, noise=noise_concrete)  # type: ignore [misc]
+            psi_init = density_mat(random_state(n_qubits))
+            values = {param: torch.rand(batch_size) for param in params}
+            psi_expected = op_concrete(psi_init, values=values)
+            psi_star = op_concrete_noise(psi_init, values=values)
+            assert torch.allclose(psi_star, psi_expected, rtol=RTOL, atol=ATOL)
```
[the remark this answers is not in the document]
AprilSweettooth: I think this is because there are things deeper in the code that I couldn't understand, with the quantum_operation stuff and target

chMoussa:
> I've tried hatch env prune but it does not work, but thanks for the suggestion

Hey @AprilSweettooth , sorry it did not work. Another team did not have this problem and managed to solve the issue before.

chMoussa:
Closing as issue #329 was completed in another MR.

Issue #284, opened 2024-10-10
[Feature] Support `ConcretizedCallable` as parameter type for parametric gates with adjoint differentiation mode 


vytautas-a:
Support `adjoint` differentiation mode for parametric gates where the differentiable parameter is given as `ConcretizedCallable`.

mlahariya:
@vytautas-a, Can you provide a bit more description on this?

Also, what could be a possible solution - should we keep the adjoint disabled for ConcretizedCallable? implement it seperately?

RolandMacDoland:
@vytautas-a Is it still relevant ? I had the feeling we weren't using `ConcretizedCallable` anymore.

chMoussa:
> [@vytautas-a](https://github.com/vytautas-a) Is it still relevant ? I had the feeling we weren't using `ConcretizedCallable` anymore.

@RolandMacDoland Yes we are still using it. So we can still consider it relevant.

chMoussa:
> [@vytautas-a](https://github.com/vytautas-a), Can you provide a bit more description on this?
> 
> Also, what could be a possible solution - should we keep the adjoint disabled for ConcretizedCallable? implement it seperately?

Hey @mlahariya, no we should extend the adjoint with `ConcretizedCallable` support. I added extra context to @vytautas-a 's issue.

Erik-Chagas:
Working on it!

Erik-Chagas:
Hi @chMoussa @vytautas-a, do we need to add a new test in ``tests/differentiation/test_differentiation.py`` or change an existing test?

chMoussa:
> Hi @chMoussa @vytautas-a, do we need to add a new test in ``tests/differentiation/test_differentiation.py`` or change an existing test?

Probably a new one.

Erik-Chagas:
I have a question. Would the main task be to implement embedding for diff mode adjoint?

chMoussa:
> I have a question. Would the main task be to implement embedding for diff mode adjoint?

@Erik-Chagas The task would be to indeed extend diff mode adjoint with embedding and adding tests for testing it out against the `AD` diff mode.

chMoussa:
> @chMoussa, does this feature need to be supported without using Embedding?

Hello @neogyk , I think it mostly needs the ConcretizedCallable. What would be an example without Embedding in this case?

chMoussa:
> [@chMoussa](https://github.com/chMoussa). The case without embedding will be:
> 
> ```
[...]

PR #341, opened 2025-06-11
Added support for Concretized Callable for parametric gate

neogyk:
#284 Issue, supporting the Concretized Callable as a parameter for the gate in the circuit and adding the ability to calculate the gradient.

### DIFF
--- a/pyqtorch/differentiation/adjoint.py
+++ b/pyqtorch/differentiation/adjoint.py
@@ -54,7 +54,7 @@ class AdjointExpectation(Function):
         state: Tensor,
         observable: Observable,
         embedding: Embedding | None,
-        param_names: list[str],
+        param_names: list[str | ConcretizedCallable],
         *param_values: Tensor,
     ) -> Tensor:
         ctx.circuit = circuit
@@ -68,11 +68,17 @@ class AdjointExpectation(Function):
         values = param_dict(param_names, param_values)
         ctx.out_state = circuit.run(state, values, embedding)
         ctx.projected_state = observable(ctx.out_state, values)
+
         for op in ctx.circuit.flatten()[::-1]:
-            if isinstance(op.param_name, ConcretizedCallable):
-                if op.is_parametric and op.param_name not in ctx.param_names:
+            if op.is_parametric:
+                if (
+                    isinstance(op.param_name, ConcretizedCallable)
+                    and op.param_name not in ctx.param_names
+                ):
                     torch.set_grad_enabled(True)
-                    values[op.param_name] = op.param_name(values).requires_grad_()
+                    values[op.param_name] = op.param_name.evaluate(
+                        values
+                    ).requires_grad_()
                     ctx.param_names.append(op.param_name)
                 torch.set_grad_enabled(False)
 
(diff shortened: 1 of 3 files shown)


chMoussa:
Hello @neogyk ,

You'll probably need:
- merge the latest changes from the base branch,
- run `pre-commit run --all-files` for linting before pushing
- add a test then make sure tests pass.
- Remove debugging traces.

pyqtorch/differentiation/adjoint.py
```diff
@@ -58,45 +60,80 @@ def forward(
...
-        ctx.save_for_backward(*param_values)
+        for op in ctx.circuit.flatten()[::-1]:
+            if isinstance(op.param_name, ConcretizedCallable):
+                if op.is_parametric and op.param_name not in ctx.param_names:
+                    torch.set_grad_enabled(True)
+                    values[op.param_name] = op.param_name(values).requires_grad_()
+                    ctx.param_names.append(op.param_name)
+                torch.set_grad_enabled(False)
+
+        ctx.save_for_backward(*values.values())
```
chMoussa, line 79: Hm I am not sure the forward method has to change. I would say only the backward should. A user could also decide to not differentiate wrt `ConcretizedCallable`, so one should check in the backward if requires_grad is True.

pyqtorch/differentiation/adjoint.py
```diff
@@ -58,45 +60,80 @@ def forward(
...
-                            op.jacobian(values, ctx.embedding),
-                            op.qubit_support,
-                        )
-                        grad = grad_out * 2 * inner_prod(ctx.projected_state, mu).real
-                    if grads_dict[param_name] is not None:
-                        grads_dict[param_name] += grad
-                    else:
-                        grads_dict[param_name] = grad
 
+                    def calculate_grad(argname):
```
chMoussa, line 105: function needs a typing

pyqtorch/differentiation/adjoint.py
```diff
@@ -58,45 +60,80 @@ def forward(
...
+
                 ctx.out_state = apply_operator(
                     ctx.out_state, op.dagger(values, ctx.embedding), op.qubit_support
                 )
                 if (
                     isinstance(op, (Parametric, HamiltonianEvolution))
                     and op.is_parametric
                 ):
+                    if isinstance(op.param_name, ConcretizedCallable):
+                        grads_dict[op.param_name] = None
```
chMoussa, line 101: so not sure I follow but why doing so? Feel like again one should just check if the parameters in ConcretizedCallable are among the `grads_dict` keys.

pyqtorch/differentiation/adjoint.py
```diff
@@ -58,45 +60,80 @@ def forward(
...
+                        for argname in op.param_name.independent_args:
+                            grads_dict[argname] = deepcopy(grad)
+                            if grads_dict[argname] is not None:
+                                grads_dict[argname] *= deepcopy(
+                                    torch.autograd.grad(
+                                        values[op.param_name],
+                                        values[argname],
+                                        retain_graph=True,
+                                    )[0]
+                                )
```
chMoussa, line 134: Could you explain what goes on here? Especially why multiplying deepcopies of grads?

chMoussa (commented):
Hello @neogyk , I have a few questions. Note that tests and docs fail in CI. If you use hatch, you can test via:
`hatch -e tests run test
``hatch run docs:serve`

chMoussa:
Hello @neogyk , this is the last day to finish it. Will you be able manage?
Don't forget to pre-commit, fix tests and make sure the doc can be generated without error.

chMoussa:
> @chMoussa, could you please review the changes?

@neogyk could you run precommit?

chMoussa (approved):
@neogyk Since time is limited and this solution seems to work, I will approve it. I will optimize it further in a later PR. Thanks for your contribution!

